PR #27, opened 2018-08-21
WinML Dashboard

tiagoshibata:


Tools/WinML-Dashboard/public/requirements.txt
```diff
@@ -0,0 +1,17 @@
+--extra-index-url https://tiagoshibata.github.io/PyPI/simple
```
zhangxiang1993: Could you replace this with a microsoft page rather than a personal one.

tiagoshibata: Yes, you'll have to fork [tiagoshibata/PyPI](https://github.com/tiagoshibata/PyPI) to a Microsoft org and you should be all set

tiagoshibata: The PyPI repo and web interface are created with [dumb-pypi](https://github.com/chriskuehl/dumb-pypi)

PR #39, opened 2018-09-12
Fixing binding CSV as input in WriteDataToBinding.

ryanlai2:
Data in the input buffer wasn't being populated correctly in WriteDataToBinding. Fixed so that the value of the data is being written to the pointer address.

Also, the output for  TensorKind::Float was made more useful by outputting the max value in the result vector as well as the index.

### DIFF
--- a/Tools/WinMLRunner/BindingUtilities.h
+++ b/Tools/WinMLRunner/BindingUtilities.h
@@ -186,7 +186,7 @@ namespace BindingUtilities
         {
             T value;
             std::stringstream(elementString) >> value;
-            data = &value;
+            *data = value;
             data++;
         }
     }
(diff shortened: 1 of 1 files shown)


Tools/WinMLRunner/BindingUtilities.h
```diff
@@ -186,7 +186,7 @@ namespace BindingUtilities
         {
             T value;
             std::stringstream(elementString) >> value;
-            data = &value;
+            *data = value;
```
walrusmcd: great catch !

Tools/WinMLRunner/BindingUtilities.h
```diff
@@ -403,8 +403,19 @@ namespace BindingUtilities
                 case TensorKind::Float:
                 {
                     auto resultVector = results.Lookup(desc.Name()).as<TensorFloat>().GetAsVectorView();
-                    auto output = resultVector.GetAt(0);
-                    std::wcout << " Result: " << output << std::endl;
+                    UINT maxIndex = 0;
```
walrusmcd: looks good.   comment for the future : it looks like this function is basically trying to create a "printable" form of a tensor.   kind of cool, but like you saw with this one, it's not always a good solution to dump index[0] of the tensor.   This code will work for classification models that return a vector of floats, as percentages of predictions.  printing the max is good here.   But I think in the future we might want a more general purpose way of printing or dumping out the output tensors.

ryanlai2: Yep thanks for the idea! I'm going to add work items to the back log where it'll be nice to output entire result vector into a form that is useful (csv format, image, etc)

PR #41, opened 2018-09-14
Throw nonzero return code if load / bind / evaluate fails.

ryanlai2:
This change refines the error catching and the console messages that are displayed if a code path throws an error when loading / binding / evaluating a model.

This change also refines existing error messages.

The error codes that are being returned from main are the codes extracted from the HRESULT errors being thrown from the WinML API.


ryanlai2:
This change also adds CSV versions of the fish and kitten images

PR #42, opened 2018-09-17
Add Squeezenet Test and set up for running samples test on azure devops

youngkim93:
This change includes Squeezenet Samples test (CS project) using kitten image. 
Both squeezenet test and mnist test will be run with vstest with MSTest adapter.


Testing/SamplesTest/SqueezenetTest/SqueezenetTest.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        [TestMethod]
+        public void TestTabbyCat()
+        {
+            string tabbyCatPath = System.AppDomain.CurrentDomain.BaseDirectory + "\\kitten_224.png";
+            resetButton.Click();
+            System.Threading.Thread.Sleep(100);
+            loadModelButton.Click();
+            // wait for model to load
+            System.Threading.Thread.Sleep(2000);
```
ryanlai2: There doesn't seem to be any way for the UI test to know when the model is finished loading. So after discussing, sleep being used here is fine.

PR #45, opened 2018-09-18
Changed order so that cpu path executes before gpu path

ryanlai2:
When running evaluate on both CPU and GPU by default without specifying a flag, run evaluate on CPU first so it can show work for loading a model on the first run.


zhangxiang1993:
Refactor the output on the console. And remove load time inside the evaluation.

PR #51, opened 2018-09-21
More Samples Tests

youngkim93:
- Add test cases for style transfer (different styles using default image)
- Squeeznet sample: Add tinca tinca inference test
- MNIST CPPCX sample test


Testing/SamplesTest/SamplesTest/StyleTransferTest.cs
```diff
@@ -0,0 +1,170 @@
...
+        public void TestStyle(Style style)
+        {
+            styleElements[style].Click();
+            var cts = new CancellationTokenSource();
+
+            var styleTransferTask = Task.Factory.StartNew(state =>
+            {
+                styleElements[style].Click();
+                var token = (CancellationToken)state;
+                while (statusBlock.Text != "Done!")
```
smk2007: while (statusBlock.Text != "Done!" && stopwatch.ElapsedMilliseconds < styleTransferTimeout)
{
    Thread.Sleep(500);
}

PR #57, opened 2018-10-09
Added command line arguments to account for mean/std dev and scale on…

pmbrown1055:
… input images to allow for result validation.  Cleaned up some performance measurement issues (removed 'SUCCESS' print from inner loop).  Added CPU memory measurement to GPU runs for run comparison.

### DIFF
--- a/Tools/WinMLRunner/CommandLineArgs.h
+++ b/Tools/WinMLRunner/CommandLineArgs.h
@@ -17,6 +17,9 @@ public:
     const std::wstring& ImagePath() const { return m_imagePath; }
     const std::wstring& CsvPath() const { return m_csvData; }
 
+	const float Scale() const { return m_scale; }
+	const std::array<float, 3>& MeanStdDev() const { return m_meanStdDev; }
+
     const std::wstring& FolderPath() const { return m_modelFolderPath; }
     const std::wstring& ModelPath() const { return m_modelPath; }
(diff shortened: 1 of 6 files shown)


ryanlai2:
Hey Paul! Would it be possible for you to write some tests for this change? Are you able to run the existing winmlrunner tests in Visual Studio?

Tools/WinMLRunner/CommandLineArgs.h
```diff
@@ -34,6 +37,9 @@ class CommandLineArgs
     std::wstring m_modelFolderPath;
     std::wstring m_modelPath;
     std::wstring m_imagePath;
+
+	float m_scale;
```
ryanlai2: This scale member variable needs to be initialized. In line 339 in BindingUtilities.h, you check if m_scale!=1.0
```suggestion
	float m_scale = 1.0;
```

ryanlai2:
Hey Paul! Just checking in with you, would you be able to provide some more information on the expected behavior of your image scaling changes? Is there some input parameters I can pass into winmlrunner.exe and look at the expected output? Also, there have been some big changes since your PR so a merge may be needed to update this PR.

Thanks!

pmbrown1055:
There are some more changes we are working on to generalize this change, and general enhancements.  Will also come back with unit tests for the feature that are implemented, and the other initialization fixed that you mentioned.

Thanks!

PR #59, opened 2018-10-10
Add more configurations to gather better perf metrics

PatriceVignola:
@Microsoft/winml @ryanlai2

This is a big PR for the following reasons:

- A lot of the diff is simply refactored code that was needed to implement the following 12 permutations: CPUDevice/GPUDevice, CPUBinding/GPUBinding, RGB/BGR/Tensor. I didn't remove any existing command line flags and the defaults are still the same, but users can now get more precise perf data based on the exact hardware/input/binding they need to profile.

- Because of all the new perf metrics, I also had to change the format of the CSV (otherwise we would have needed at least 50 new columns, and must of them would contain N/A data). Instead of having all the configurations for a model on the same line, we now have a different line for every unique configuration.

- I fixed some typos I stumbled upon and some formatting issues (e.g. Ouput > Output).

- I also added tests for garbage input and improved the existing tests by making sure that the output CSVs contain the right number of lines. Also refactored tests to make it easier and less error-prone to build a command.


Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
```diff
@@ -25,57 +28,285 @@ static int RunProc(LPWSTR commandLine)
 
 namespace WinMLRunnerTest
 {
+    static const std::wstring CURRENT_PATH = FileHelper::GetModulePath();
```
PatriceVignola: Added a bunch of new tests and refactored old ones.

Testing/WinMLRunnerTest/WinMLRunnerTest.vcxproj
```diff
@@ -204,6 +204,17 @@
       <DeploymentContent Condition="'$(Configuration)|$(Platform)'=='Release|x64'">true</DeploymentContent>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="..\..\SharedContent\models\Add_ImageNet224.onnx">
```
PatriceVignola: I found out that the Add_ImageNet1920.onnx file asked in one test was not even in the repo, so the test wasn't really accurate. I changed the test to use Add_ImageNet224.onnx since I couldn't find the original

Tools/WinMLRunner/BindingUtilities.h
```diff
@@ -1,139 +1,58 @@
...
 using namespace Windows::Storage;
 using namespace winrt::Windows::AI::MachineLearning;
+using namespace winrt::Windows::Foundation;
+using namespace winrt::Windows::Foundation::Collections;
+    using namespace winrt::Windows::Graphics::DirectX;
+    using namespace winrt::Windows::Graphics::Imaging;
 
 namespace BindingUtilities
 {
-    void BindTensorsFromGarbageData(LearningModelBinding context, LearningModel model)
```
PatriceVignola: Bing refactoring here. Instead of using this huge switch statement at 2 different places, I place it in a single method called by both garbage data and CSV data.

Tools/WinMLRunner/Filehelper.cpp
```diff
@@ -0,0 +1,25 @@
+#include "Filehelper.h"
```
PatriceVignola: This is used by dllload.cpp

Tools/WinMLRunner/Filehelper.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
```
PatriceVignola: This is used by dllload.cpp.

Tools/WinMLRunner/README.md
```diff
@@ -17,26 +17,41 @@ You must unzip the entire archive if you intend to build the samples.
...
 Required command-Line arguments:
--model <path>         : Fully qualified path to a .onnx or .pb model file.
+-model <path>            : Fully qualified path to a .onnx or .pb model file.
       or
--folder <path>        : Fully qualifed path to a folder with .onnx and/or .pb models, will run all of the models in the folder.
+-folder <path>           : Fully qualifed path to a folder with .onnx and/or .pb models, will run all of the models in the folder.
 
 #Optional command-line arguments:
 -perf                    : Captures GPU, CPU, and wall-clock time measurements. 
 -iterations <int>	     : Number of times to evaluate the model when capturing performance measurements.
```
PatriceVignola: Alignment (Tabs -> Spaces)

Tools/WinMLRunner/dllload.cpp
```diff
@@ -0,0 +1,73 @@
+#include "FileHelper.h"
```
PatriceVignola: This is almost the same file as [this gist](https://gist.github.com/kennykerr/3649b0878824ead2755e1f7c86e4270a), with some modifications specific to winml. This code makes it possible to use a local Windows.AI.MachineLearning.dll DLL instead of the System32 one.

Tools/WinMLRunner/OutputHelper.h
```diff
@@ -9,65 +8,162 @@
...
+            if (SUCCEEDED(adapter->GetDesc(&description)))
+            {
+                std::wcout << L"GPU: " << description.Description << std::endl;
+                std::cout << std::endl;
+            }
+        }
     }
 
-    void PrintGPUTimes(Profiler<WINML_MODEL_TEST_PERF> &profiler, UINT iterations)
+    void PrintResults(const Profiler<WINML_MODEL_TEST_PERF> &profiler, uint32_t numIterations, DeviceType deviceType, InputBindingType inputBindingType, InputDataType inputDataType) const
```
PatriceVignola: We now have a single PrintResults() method, no matter which configuration is used.

Tools/WinMLRunner/OutputHelper.h
```diff
@@ -125,23 +221,6 @@ class OutputHelper
         }
     }
 
-    static void PrintFeatureDescriptorInfo(const winrt::Windows::AI::MachineLearning::ILearningModelFeatureDescriptor &descriptor)
```
PatriceVignola: This has been moved somewhere else.

Tools/WinMLRunner/OutputHelper.h
```diff
@@ -184,66 +263,40 @@ class OutputHelper
         return false;
     }
 
-    void PrintModelInfo(std::wstring modelPath, LearningModel model)
```
PatriceVignola: This has been moved somewhere else.

Tools/WinMLRunner/OutputHelper.h
```diff
@@ -265,98 +318,66 @@ class OutputHelper
...
-         m_GPUEvalSharedMemoryUsage = 0;
-         m_GPUEvalDedicatedMemoryUsage = 0;
-
-         m_CPULoadTime = 0;
-         m_CPUBindTime = 0;
-         m_CPUEvalTime = 0;
-         m_CPUEvalMemoryUsage = 0;
-
-
+         m_clockEvalTime = 0;
```
PatriceVignola: Ideally, we shouldn't keep any member variable in the output helper (especially when we modify them in the Print methods), but this is a good start.

PR #69, opened 2018-10-22
Modify readme with new perf features

PatriceVignola:


Tools/WinMLRunner/Main.cpp
```diff
@@ -296,12 +296,6 @@ HRESULT EvaluateModels(
             {
                 for (auto inputDataType : inputDataTypes)
                 {
-                    if (args.PerfCapture())
```
PatriceVignola: The perf data was getting cleared before we could even write the load time, so I placed it at the end of the loop instead.

PatriceVignola:
@Microsoft/winml @ryanlai2

PR #70, opened 2018-10-24
Add CreateDeviceOnClient and CreateDeviceInWinML flags to WinMLRunner

PatriceVignola:
@Microsoft/winml @ryanlai2 

I stumbled upon another configuration that can change the runtime: device create on the client and passed to WinML versus device created inside WinML. When the client creates the device himself and uses this device to create a video frame, and creates a session with this device, the binding will be faster on the GPU since we avoid a cross-device and cross-GPU copy.

So I added these 2 new flags, which bump the total number of configurations to 24. Even if the model is running on the CPU, creating the device on the client side can have an impact since its input might be GPU bound and will have to be tensorized. I updated the README with the new flags.

I also changed the garbage image generation to make it deterministic between runs. This will be a lot easier to debug than seeds based on the clock and it will be possible to write validation tests by using the same seeds.

### DIFF
--- a/Tools/WinMLRunner/BindingUtilities.h
+++ b/Tools/WinMLRunner/BindingUtilities.h
@@ -5,16 +5,19 @@
 #include "ModelBinding.h"
 #include "CommandLineArgs.h"
 
-using namespace Windows::Media;
-using namespace Windows::Storage;
+using namespace winrt::Windows::Media;
+using namespace winrt::Windows::Storage;
 using namespace winrt::Windows::AI::MachineLearning;
-using namespace winrt::Windows::Foundation;
 using namespace winrt::Windows::Foundation::Collections;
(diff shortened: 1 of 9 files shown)


Tools/WinMLRunner/BindingUtilities.h
```diff
@@ -36,8 +39,7 @@ namespace BindingUtilities
 
         // Generate random values for the image
```
ryanlai2: This won't be random values anymore

ryanlai2:
Cool change Pat! Can you update the README.md to include some example usage?

Tools/WinMLRunner/BindingUtilities.h
PatriceVignola: Good catch! I'll change the comment.

PR #74, opened 2018-10-29
User/ryanlai2/add arm64

ryanlai2:
This PR primarily adds the ability for WinMLRunner.exe to be build with the ARM64 architecture. Unfortunately the Visual Studio tests couldn't be built with ARM64 because it isn't supported so only the winmlrunner.exe can be built.

This PR also makes the  Reset() method for OutputHelper class more explicit and it removes the reset for load time because the Reset() method was only used in Main.cpp to reset the bind and evaluation times for each evaluation iteration. However, it resets the load time as well which is undesirable because we want to preserve the load time when printing out the output information.

Another change in this PR is to statically link vcruntime140.dll and MSVCP140.dll when building for Release. These 2 DLLs are installed when Visual Studio is installed so the release version of Winmlrunner binary shouldn't rely on these DLLs to run on a machine that doesn't have VS installed.

This change also fixes the bug where winmlrunner couldn't build on x86

### DIFF
--- a/Tools/WinMLRunner/README.md
+++ b/Tools/WinMLRunner/README.md
@@ -5,8 +5,8 @@ The WinMLRunner is a command-line based tool that can run .onnx or .pb models wh
 
 ## Prerequisites
 - [Visual Studio 2017 Version 15.7.4 or Newer](https://developer.microsoft.com/en-us/windows/downloads)
-- [Windows 10 - Build 17738 or higher](https://www.microsoft.com/en-us/software-download/windowsinsiderpreviewiso)
-- [Windows SDK - Build 17738 or higher](https://www.microsoft.com/en-us/software-download/windowsinsiderpreviewSDK)
+- [Windows 10 - Build 17763 or higher](https://www.microsoft.com/en-us/software-download/windowsinsiderpreviewiso)
+- [Windows SDK - Build 17763 or higher](https://www.microsoft.com/en-us/software-download/windowsinsiderpreviewSDK)
 
 ## Build the tool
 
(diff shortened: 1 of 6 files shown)


Tools/WinMLRunner/README.md
```diff
@@ -5,8 +5,8 @@ The WinMLRunner is a command-line based tool that can run .onnx or .pb models wh
 
 ## Prerequisites
 - [Visual Studio 2017 Version 15.7.4 or Newer](https://developer.microsoft.com/en-us/windows/downloads)
-- [Windows 10 - Build 17738 or higher](https://www.microsoft.com/en-us/software-download/windowsinsiderpreviewiso)
-- [Windows SDK - Build 17738 or higher](https://www.microsoft.com/en-us/software-download/windowsinsiderpreviewSDK)
+- [Windows 10 - Build 17763 or higher](https://www.microsoft.com/en-us/software-download/windowsinsiderpreviewiso)
+- [Windows SDK - Build 17763 or higher](https://www.microsoft.com/en-us/software-download/windowsinsiderpreviewSDK)
```
PatriceVignola: Good catch, it's hard to keep the documentation up-to-date sometimes :)

ryanlai2: Yep yep I agree! I think we'll need a large refactor of the documentation sometime in the future too

Tools/WinMLRunner/WinMLRunner.vcxproj
```diff
@@ -140,15 +199,36 @@
       <SDLCheck>true</SDLCheck>
       <ShowIncludes>true</ShowIncludes>
       <LanguageStandard>stdcpp17</LanguageStandard>
+      <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
+    </ClCompile>
+    <Link>
+      <EnableCOMDATFolding>true</EnableCOMDATFolding>
+      <OptimizeReferences>true</OptimizeReferences>
+      <SubSystem>Console</SubSystem>
+      <AdditionalDependencies>dxgi.lib;d3d12.lib;winml.lib;windowsapp.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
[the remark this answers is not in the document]
ryanlai2: Fixed!

PR #81, opened 2018-11-02
Added Emoji8 source code to Samples folder

mccoykillian:


Samples/Emoji8/UWP/cs/Emoji8/Services/IntelligenceService.cs
```diff
@@ -0,0 +1,297 @@
...
+            try
+            {
+                _model = await LearningModel.LoadFromStorageFileAsync(modelStorageFile);
+            }
+            catch(Exception ex)
+            {
+                Debug.WriteLine(ex.Message);
+            }
+
+            _device = new LearningModelDevice(LearningModelDeviceKind.Default);
```
martinb35: if you're just going to use the default device, you actually don't need to specify it - you can just create a session without the device parameter: "_session = new LearningModelSession(_model);".
alternatively, you could try to run this on the GPU too if the user has one.

Samples/Emoji8/UWP/cs/Emoji8/Pages/EmotionPage.xaml.cs
```diff
@@ -0,0 +1,310 @@
...
+    /// <summary>
+    /// An empty page that can be used on its own or navigated to within a Frame.
+    /// </summary>
+    public sealed partial class EmotionPage : Page
+    {
+        private static List<Emoji> _resultsEmojis { get; set; }
+        private DispatcherTimer EmojiTimer;
+        private int countdownNum;
+        private int numEmojisLeftToPerform = Constants.EMOJI_NUMBER_TO_DISPLAY;
+        private double nextPercentage;
```
youngkim93: Can we have a default value for this for consistency?

youngkim93: same for countdownNum :)

Samples/Emoji8/UWP/cs/Emoji8/Pages/EmotionPage.xaml.cs
```diff
@@ -0,0 +1,310 @@
...
+            }
+            else
+            {
+                UpdateCountdownRing();
+            }
+
+            if (countdownNum % Constants.DIVISIBLE_FACTOR == 0)
+            {
+                //display pattern 5...4...3...2...1...5... [...] ...2...1...0
+                if ((numEmojisLeftToPerform > -1 && countdownNum != 0) ||
```
youngkim93: I haven't read through all the details of this, but could numEmojisLeftToPerform be negative? By looking at the variable name it seems to me like it should never be negative :)

PR #86, opened 2018-11-08
Autoscale input images to match model input.

pmbrown1055:
Adding automatic scaling of input images to match model input dimensions.


Tools/WinMLRunner/BindingUtilities.h
```diff
@@ -65,10 +71,29 @@ namespace BindingUtilities
...
+
+            // If input dimensions are different from tensor input, then scale / crop while reading
+            if (decoder.PixelHeight() != height ||
+                decoder.PixelWidth() != width)
+            {
+                // Create a transform object with default parameters (no transform)
+                auto transform = BitmapTransform();
+                transform.ScaledHeight(height);
+                transform.ScaledWidth(width);
+                transform.InterpolationMode(BitmapInterpolationMode::Cubic);
```
ryanlai2: Is it a good idea to hardcode this interpolation mode in the transform? Would it be scenario dependent?

ryanlai2:
Thank you for the contribution! Would it be possible to have this feature as a flag so that the user can specify if they want autoscaling?

Tools/WinMLRunner/BindingUtilities.h
pmbrown1055: Possibly, my best attempt to get the model the best possible input.  Could make it part of the flag?

pmbrown1055:
> Thank you for the contribution! Would it be possible to have this feature as a flag so that the user can specify if they want autoscaling?

Would I default this to on?  It works around some functional issues.  Also, if the interpolation mode is programmable at run time, default it to Cubic and allow the user to change with another command line option, or the same one?

Tools/WinMLRunner/BindingUtilities.h
ryanlai2: Yeah let's make it an argument that's required when passing in the flag. Sounds good?

ryanlai2:
Can we have the interpolation as an argument that's required when passing in the flag? That way no defaults are necessary.

 I understand that this works around the functional issue that there wouldn't be an error thrown when an image is passed in and it has different dimensions from what the model expects. However, this error should be the expected behavior when using WinML's API. 

Adding an option to autoscale would allow the user to work around this problem, but I don't think it should be the default. Thoughts?

pmbrown1055:
Commit update per conversation comments.  Please review.

ryanlai2:
Looks good to me! Can you update the README.md with what you've added and if possible have a test that tests for these scenarios:

run winmlrunner with model with unscaled image and with no autoscale flag = verify failed and throws error
run winmlrunner with model with unscaled image and with flag set for autoscale Cubic = verify success

pmbrown1055:
Two main changes:
1. Same as before, but updated Readme.md with new option.
2. Added fish_112.png to test autoscaling.  Updated already existing test (ProvidedImageBadBinding to use new fish_112.png and return the same results as previously.)  Added AutoScaleImage to return a passing result with -autoScale Cubic option.
All tests pass.

PR #89, opened 2018-11-13
D3D12 command queue sample

orilevari:
Sample which allows the user to select from an enumerated list of device adapters, and then constructs a LearningModelDevice from a D3D12 command queue .


Samples/AdapterSelection/AdapterSelection/cpp/main.cpp
```diff
@@ -0,0 +1,239 @@
...
+	LearningModelSession session(model);
+	LearningModelBinding binding(session);
+
+	// load the image
+	printf("Loading the image...\n");
+	auto imageFrame = LoadImageFile(imagePath);
+
+	// bind the input image
+	printf("Binding...\n");
+	binding.Bind(L"data_0", ImageFeatureValue::CreateFromVideoFrame(imageFrame));
```
ryanlai2: The input name of the model shouldn't be hardcoded. It may be different for each model. You can get the input name at runtime from the model's InputFeatureDescriptors

Samples/AdapterSelection/AdapterSelection/cpp/main.cpp
```diff
@@ -0,0 +1,239 @@
...
+	// temp: bind the output (we don't support unbound outputs yet)
+	vector<int64_t> shape({ 1, 1000, 1, 1 });
+	binding.Bind(L"softmaxout_1", TensorFloat::Create(shape));
```
ryanlai2: Same thing for output. You can get the name at runtime from the model's output feature descriptors.

Samples/AdapterSelection/AdapterSelection/cpp/main.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+	// load the model
+	printf("Loading modelfile '%ws' on the selected device\n", modelPath.c_str());
+	DWORD ticks = GetTickCount();
+	auto model = LearningModel::LoadFromFilePath(modelPath);
+	ticks = GetTickCount() - ticks;
+	printf("model file loaded in %d ticks\n", ticks);
+
+	// now create a session and binding
+	//LearningModelSession session(model, spLearningDevice.as<LearningModelDevice>());
```
ryanlai2: Nit: Remove commented out code

Samples/AdapterSelection/AdapterSelection/cpp/main.cpp
```diff
@@ -0,0 +1,239 @@
...
+	LearningModelSession session(model);
```
ryanlai2: How are you using the learning model device you created? spLearningDevice?

orilevari: whoops, good catch!

Samples/AdapterSelection/AdapterSelection/cpp/main.cpp
orilevari: I also changed the output results to be extracted using model.OutputFeatures()

PR #90, opened 2018-11-13
Save Per-Iteration Performance Results

laks0209:
* Command line argument -saveiter dumps per-iteration values to csv files
* The following per-iteration values are saved in PerIterationValues.csv
   - ModelName, ImageName
   - Final Result, Hash of Output Tensor
   - CPU & GPU Memory(Diff & Start)
   - Load, Bind & Evaluate Time
* Result[FullOutputTensor].csv contains the per-iteration output tensor result

### DIFF
--- a/Tools/WinMLRunner/CommandLineArgs.cpp
+++ b/Tools/WinMLRunner/CommandLineArgs.cpp
@@ -28,6 +28,7 @@ void CommandLineArgs::PrintUsage() {
     std::cout << "  -output <fully qualified path>: csv file to write the perf results to" << std::endl;
     std::cout << "  -IgnoreFirstRun : ignore the first run in the perf results when calculating the average" << std::endl;
     std::cout << "  -silent: only errors are printed to the console" << std::endl;
+    std::cout << "  -savePerIterationPerf : save per iteration performance results to csv file" << std::endl;
     std::cout << "  -debug: print trace logs" << std::endl;
     std::cout << "  -autoScale <interpolationMode>: Enable image autoscaling and set the interpolation mode [Nearest, Linear, Cubic, Fant]" << std::endl;
 }
@@ -118,6 +119,10 @@ CommandLineArgs::CommandLineArgs()
         else if ((_wcsicmp(args[i], L"-silent") == 0))
         {
             m_silent = true;
+        }
+        else if ((_wcsicmp(args[i], L"-savePerIterationPerf") == 0))
+        {
+            m_perIterCapture = true;
         }
         else if ((_wcsicmp(args[i], L"-autoScale") == 0) && (i + 1 < numArgs))
         {
--- a/Tools/WinMLRunner/Main.cpp
+++ b/Tools/WinMLRunner/Main.cpp
@@ -11,7 +11,7 @@ Profiler<WINML_MODEL_TEST_PERF> g_Profiler;
 
 using namespace winrt::Windows::Graphics::DirectX::Direct3D11;
 
-LearningModel LoadModel(const std::wstring path, bool capturePerf, bool silent, OutputHelper& output)
+LearningModel LoadModel(const std::wstring path, bool capturePerf, OutputHelper& output, const CommandLineArgs& args, uint32_t iterationNum)
 {
     LearningModel model = nullptr;
     output.PrintLoadingInfo(path);
@@ -27,6 +27,10 @@ LearningModel LoadModel(const std::wstring path, bool capturePerf, bool silent, 
         if (capturePerf)
         {
             WINML_PROFILING_STOP(g_Profiler, WINML_MODEL_TEST_PERF::LOAD_MODEL);
+            if (args.PerIterCapture())
+            {
+                output.SaveLoadTimes(g_Profiler, iterationNum);
+            }
         }
     }
     catch (hresult_error hr)
@@ -93,7 +97,7 @@ std::vector<ILearningModelFeatureValue> GenerateInputFeatures(
     return inputFeatures;
 }
 
-HRESULT BindInputFeatures(const LearningModel& model, const LearningModelBinding& context, const std::vector<ILearningModelFeatureValue>& inputFeatures, const CommandLineArgs& args, OutputHelper& output, bool capturePerf)
+HRESULT BindInputFeatures(const LearningModel& model, const LearningModelBinding& context, const std::vector<ILearningModelFeatureValue>& inputFeatures, const CommandLineArgs& args, OutputHelper& output, bool capturePerf, uint32_t iterationNum)
 {
     assert(model.InputFeatures().Size() == inputFeatures.size());
 
@@ -115,6 +119,10 @@ HRESULT BindInputFeatures(const LearningModel& model, const LearningModelBinding
         if (capturePerf)
         {
             WINML_PROFILING_STOP(g_Profiler, WINML_MODEL_TEST_PERF::BIND_VALUE);
+            if (args.PerIterCapture())
+            {
+                output.SaveBindTimes(g_Profiler, iterationNum);
+            }
         }
 
         if (!args.Silent())
@@ -139,7 +147,8 @@ HRESULT EvaluateModel(
     bool isGarbageData,
     const CommandLineArgs& args,
     OutputHelper& output,
-    bool capturePerf
+    bool capturePerf,
+    uint32_t iterationNum
 )
 {
     LearningModelEvaluationResult result = nullptr;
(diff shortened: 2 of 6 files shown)


Tools/WinMLRunner/OutputHelper.h
```diff
@@ -354,7 +407,9 @@ class OutputHelper
                      << "Wall-clock Load (ms)" << ","
                      << "Wall-clock Bind (ms)" << ","
                      << "Wall-clock Evaluate (ms)" << ","
-                     << "Wall-clock total time (ms)" << std::endl;
+                     << "Wall-clock total time (ms)" << ","
+                     << "PerIterationFile" << ","
```
ryanlai2: For users that primarily use -perf flag to capture performance and do not care about the per iterations and the results file, we don't want this in their file because it will mostly be blank.

ryanlai2:
May I request this PR to exclude changes from Paul Brown's autoscaling changes so that it's easier to review?? https://github.com/Microsoft/Windows-Machine-Learning/pull/86

Tools/WinMLRunner/OutputHelper.h
```diff
@@ -273,6 +283,49 @@ class OutputHelper
         return false;
     }
 
+    void SaveEvalTimes(Profiler<WINML_MODEL_TEST_PERF> &profiler, uint32_t IterNum)
+    {
+        m_CPUWorkingDiff[IterNum] = profiler[EVAL_MODEL].GetCpuWorkingDiff();
```
ryanlai2: If we evaluate twice, once on the CPU and once on the GPU, m_CPUWorkingDiff, m_CPUWorkingStart, m_GPUSharedDiff, m_GPUSharedStart, and m_GPUDedicatedDiff vectors are all of size 0. There is an exception that is thrown because we are trying to write to it here.

Tools/WinMLRunner/OutputHelper.h
```diff
@@ -374,8 +429,179 @@ class OutputHelper
...
+            
+            if (bNewFile)
+            {
+                fout << "Model Name" << ","
+                     << "Image Name" << ","
+                     << "Iterations" << ","
+                     << "Iteration Number " << ","
+                     << "Result" << ","
+                     << "Hash" << ","
+                     << "CPU Working Set Diff" << ","
```
ryanlai2: Is this working set diff in MBs?

ryanlai2:
Per offline discussion with @pmbrown1055 , let's separate this PR into 2 PRs. One for outputting per-iteration performance results and one for  outputting tensors into a CSV format/.

laks0209:
The pull request has been updated as per previous conversation. The current pull request only saves the per-iteration performance results. Could you please review the changes?

Tools/WinMLRunner/OutputHelper.h
```diff
@@ -14,7 +15,15 @@ using namespace Windows::Storage::Streams;
 class OutputHelper
 {
 public:
-    OutputHelper(bool silent) : m_silent(silent) {}
+    OutputHelper(int num, bool silent)
```
ryanlai2: Can we make the naming of "num" to be more specific?

Tools/WinMLRunner/Main.cpp
```diff
@@ -146,7 +147,8 @@ HRESULT EvaluateModel(
     bool isGarbageData,
     const CommandLineArgs& args,
     OutputHelper& output,
-    bool capturePerf
+    bool capturePerf,
+    uint32_t iterNum
```
ryanlai2: can we change "iterNum" to "iterationNum"?

laks0209: Sure.

Tools/WinMLRunner/OutputHelper.h
laks0209: Okay.

Tools/WinMLRunner/Main.cpp
```diff
@@ -270,12 +276,12 @@ HRESULT EvaluateModel(
     // Add one more iteration if we ignore the first run
     uint32_t numIterations = args.NumIterations() + args.IgnoreFirstRun();
 
-    bool isGarbageData = !args.CsvPath().empty() || !args.ImagePath().empty();
+    bool isGarbageData = args.CsvPath().empty() || args.ImagePath().empty();
```
ryanlai2: Good catch! But should it be:
`args.CsvPath().empty() && args.ImagePath().empty(); `
If there is no CSV and also there is no Image provided, then we should use Garbage Data.

laks0209: Yes. That's right! I will correct it.

Tools/WinMLRunner/Main.cpp
```diff
@@ -270,12 +276,12 @@ HRESULT EvaluateModel(
...
 
     // Run the binding + evaluate multiple times and average the results
     for (uint32_t i = 0; i < numIterations; i++)
     {
-        bool captureIterationPerf = args.PerfCapture() && (!args.IgnoreFirstRun() || i > 0);
+        bool captureIterationPerf = ( args.PerfCapture() && (!args.IgnoreFirstRun() || i > 0) ) || (args.PerIterCapture());
```
ryanlai2: Nit: Inconsistent Spacing

laks0209: Corrections made in the latest commit.

Tools/WinMLRunner/OutputHelper.h
```diff
@@ -370,35 +381,82 @@ class OutputHelper
...
+                     << imgName << ","
+                     << args.NumIterations() << ","
+                     << i + 1 << ","
+                     << m_CPUWorkingDiff[i] << ","
+                     << m_CPUWorkingStart[i] << ","
+                     << m_GPUSharedDiff[i] << ","
+                     << m_GPUSharedStart[i] << ","
+                     << m_GPUDedicatedDiff[i] << ","
+                     << m_clockLoadTimes[i] << ","
+                     << m_clockBindTimes[i] << ","
```
ryanlai2: Can you merge with master? We're deprecating the clock timings because the profiler already collects the times for us

ryanlai2: You can get the bind times with: 
```
double bindTime = profiler[BIND_VALUE].GetAverage(CounterType::TIMER);
```

ryanlai2: Same things with load and evaluation times ^^

laks0209: Merged with master. Please review the new commit.

ryanlai2: m_clockLoadTimes, m_clockBindTimes, m_clockEvalTimes aren't being populated after the merge

laks0209: Thats right. Sorry, I have corrected it in the new commit.

Tools/WinMLRunner/OutputHelper.h
```diff
@@ -263,6 +275,37 @@ class OutputHelper
...
 
+    void SaveLoadTimes(Profiler<WINML_MODEL_TEST_PERF> &profiler, uint32_t iterNum)
+    {
+        m_clockLoadTimes[iterNum] = profiler[LOAD_MODEL].GetClockTime();
+    }
+    void SaveBindTimes(Profiler<WINML_MODEL_TEST_PERF> &profiler, uint32_t iterNum)
+    {
+        m_clockBindTimes[iterNum] = profiler[BIND_VALUE].GetClockTime();
+    }
+    void SaveEvalTimes(Profiler<WINML_MODEL_TEST_PERF> &profiler, uint32_t iterNum)
```
ryanlai2: Can we change the naming of the method to "SaveEvalPerformance"?

Tools/WinMLRunner/CommandLineArgs.cpp
```diff
@@ -28,6 +28,7 @@ void CommandLineArgs::PrintUsage() {
     std::cout << "  -output <fully qualified path>: csv file to write the perf results to" << std::endl;
     std::cout << "  -IgnoreFirstRun : ignore the first run in the perf results when calculating the average" << std::endl;
     std::cout << "  -silent: only errors are printed to the console" << std::endl;
+    std::cout << "  -savePeriterPerf : save per iteration performance results to csv file" << std::endl;
```
ryanlai2: Can we change this flag to "savePerIterationPerf"?

ryanlai2:
Looks good to me! I made a couple more comments, after those are fixed. Then we should be good to go :)

laks0209:
Made corrections as per the comments :-)

PR #91, opened 2018-11-16
Add input pre-process support by Scale, Mean, and Std Dev to enable c…

pmbrown1055:
…orrect output predicitons for specific CV models (e.g. densenet121, resnet50).

### DIFF
--- a/Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
+++ b/Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
@@ -6,6 +6,7 @@
 #include <Winbase.h>
 #include <filesystem>
 #include <fstream>
+#include <iostream>
 #include <algorithm>
 #include <vector>
 #include <direct.h>
@@ -32,13 +33,16 @@ static HRESULT RunProc(LPWSTR commandLine)
     return HRESULT_FROM_WIN32(exitCode);
 }
 
+// Use this test method definition, if the test needs access to the METHOD_NAME
+#define TEST_METHOD_WITH_NAME(methodName) TEST_METHOD(methodName) { const std::wstring METHOD_NAME(L#methodName);
+
 namespace WinMLRunnerTest
 {
     static const std::wstring CURRENT_PATH = FileHelper::GetModulePath();
     static const std::wstring EXE_PATH = CURRENT_PATH + L"WinMLRunner.exe";
     static const std::wstring INPUT_FOLDER_PATH = CURRENT_PATH + L"test_folder_input";
     static const std::wstring OUTPUT_PATH = CURRENT_PATH + L"test_output.csv";
-    static const std::wstring TENSOR_DATA_PATH = CURRENT_PATH + L"\\TestResult";
+    static const std::wstring TENSOR_DATA_PATH = CURRENT_PATH + L"TestResults";
 
     static std::wstring BuildCommand(std::initializer_list<std::wstring>&& arguments)
     {
@@ -226,7 +230,7 @@ namespace WinMLRunnerTest
         copyCommand += model;
         copyCommand += ' ' + std::string(INPUT_FOLDER_PATH.begin(), INPUT_FOLDER_PATH.end());
         system(copyCommand.c_str());
-            }
+    }
 } // namespace WinMLRunnerTest
 
         TEST_CLASS_CLEANUP(CleanupClass)
@@ -235,6 +239,13 @@ namespace WinMLRunnerTest
             std::string copyCommand = "rd /s /q ";
             copyCommand += std::string(INPUT_FOLDER_PATH.begin(), INPUT_FOLDER_PATH.end());
             system(copyCommand.c_str());
+            try
+            {
+                std::filesystem::remove_all(std::string(TENSOR_DATA_PATH.begin(), TENSOR_DATA_PATH.end()).c_str());
+            }
+            catch (const std::filesystem::filesystem_error &)
+            {
+            }
         }
 
         TEST_METHOD_CLEANUP(CleanupMethod)
@@ -677,45 +688,88 @@ namespace WinMLRunnerTest
                 BuildCommand({ EXE_PATH, L"-model ", modelPath, L"-input", inputPath, L"-autoScale", L"Cubic" });
             Assert::AreEqual(S_OK, RunProc((wchar_t*)command.c_str()));
         }
-        TEST_METHOD(ProvidedImageInputOnlyGpuSaveTensor)
-        {
+        TEST_METHOD_WITH_NAME(ProvidedImageInputOnlyGpuSaveTensor)
             const std::wstring modelPath = CURRENT_PATH + L"SqueezeNet.onnx";
             const std::wstring inputPath = CURRENT_PATH + L"fish.png";
+            const std::wstring tensorDataPath = TENSOR_DATA_PATH + L"\\" + METHOD_NAME;
             const std::wstring command = BuildCommand({ EXE_PATH, L"-model ", modelPath, L"-input", inputPath,
-                                                        L"-SaveTensorData", L"First", TENSOR_DATA_PATH, L"-GPU" });
+                L"-SaveTensorData", L"First", tensorDataPath, L"-GPU" });
             Assert::AreEqual(S_OK, RunProc((wchar_t*)command.c_str()));
             Assert::AreEqual(true, CompareTensors(L"OutputTensorData\\Squeezenet_fish_input_GPU.csv",
-                                                  TENSOR_DATA_PATH + L"\\softmaxout_1GpuIteration1.csv"));
+                                                  tensorDataPath + L"\\softmaxout_1GpuIteration1.csv"));
         }
-        TEST_METHOD(ProvidedImageInputOnlyCpuSaveTensor)
-        {
+        TEST_METHOD_WITH_NAME(ProvidedImageInputOnlyGpuSaveTensorPreprocessImage)
             const std::wstring modelPath = CURRENT_PATH + L"SqueezeNet.onnx";
             const std::wstring inputPath = CURRENT_PATH + L"fish.png";
+            const std::wstring tensorDataPath = TENSOR_DATA_PATH + L"\\" + METHOD_NAME;
             const std::wstring command = BuildCommand({ EXE_PATH, L"-model ", modelPath, L"-input", inputPath,
-                                                        L"-SaveTensorData", L"First", TENSOR_DATA_PATH, L"-CPU" });
+                L"-SaveTensorData", L"First", tensorDataPath, L"-GPU", L"-inputImagePreprocess 1 0 0 0", L"-BGR" });
+            Assert::AreEqual(S_OK, RunProc((wchar_t*)command.c_str()));
+            Assert::AreEqual(true, CompareTensors(L"OutputTensorData\\Squeezenet_fish_input_GPU.csv",
+                                                  tensorDataPath + L"\\softmaxout_1GpuIteration1.csv"));
+        }
+        TEST_METHOD_WITH_NAME(ProvidedImageInputOnlyCpuSaveTensor)
+            const std::wstring modelPath = CURRENT_PATH + L"SqueezeNet.onnx";
+            const std::wstring inputPath = CURRENT_PATH + L"fish.png";
+            const std::wstring tensorDataPath = TENSOR_DATA_PATH + L"\\" + METHOD_NAME;
+            const std::wstring command = BuildCommand({ EXE_PATH, L"-model ", modelPath, L"-input", inputPath,
+                                                        L"-SaveTensorData", L"First", tensorDataPath, L"-CPU" });
             Assert::AreEqual(S_OK, RunProc((wchar_t*)command.c_str()));
             Assert::AreEqual(true, CompareTensors(L"OutputTensorData\\Squeezenet_fish_input_CPU.csv",
-                                                  TENSOR_DATA_PATH + L"\\softmaxout_1CpuIteration1.csv"));
+                                                  tensorDataPath + L"\\softmaxout_1CpuIteration1.csv"));
         }
-        TEST_METHOD(ProvidedImageInputOnlyGpuSaveTensorFp16)
-        {
+        TEST_METHOD_WITH_NAME(ProvidedImageInputOnlyCpuSaveTensorPreprocessImage)
+            const std::wstring modelPath = CURRENT_PATH + L"SqueezeNet.onnx";
+            const std::wstring inputPath = CURRENT_PATH + L"fish.png";
+            const std::wstring tensorDataPath = TENSOR_DATA_PATH + L"\\" + METHOD_NAME;
+            const std::wstring command = BuildCommand({ EXE_PATH, L"-model ", modelPath, L"-input", inputPath,
+                                                        L"-SaveTensorData", L"First", tensorDataPath, L"-CPU",
+                                                        L"-inputImagePreprocess 1 0 0 0", L"-BGR" });
+            Assert::AreEqual(S_OK, RunProc((wchar_t*)command.c_str()));
+            Assert::AreEqual(true, CompareTensors(L"OutputTensorData\\Squeezenet_fish_input_CPU.csv",
+                                                  tensorDataPath + L"\\softmaxout_1CpuIteration1.csv"));
+        }
+        TEST_METHOD_WITH_NAME(ProvidedImageInputOnlyGpuSaveTensorFp16)
             const std::wstring modelPath = CURRENT_PATH + L"SqueezeNet_fp16.onnx";
             const std::wstring inputPath = CURRENT_PATH + L"fish.png";
+            const std::wstring tensorDataPath = TENSOR_DATA_PATH + L"\\" + METHOD_NAME;
             const std::wstring command = BuildCommand({ EXE_PATH, L"-model ", modelPath, L"-input", inputPath,
-                                                        L"-SaveTensorData", L"First", TENSOR_DATA_PATH, L"-GPU" });
+                                                        L"-SaveTensorData", L"First", tensorDataPath, L"-GPU" });
             Assert::AreEqual(S_OK, RunProc((wchar_t*)command.c_str()));
             Assert::AreEqual(true, CompareTensorsFP16(L"OutputTensorData\\Squeezenet_fp16_fish_input_GPU.csv",
-                                                      TENSOR_DATA_PATH + L"\\softmaxout_1GpuIteration1.csv"));
+                                                      tensorDataPath + L"\\softmaxout_1GpuIteration1.csv"));
         }
-        TEST_METHOD(ProvidedImageInputOnlyCpuSaveTensorFp16)
-        {
+        TEST_METHOD_WITH_NAME(ProvidedImageInputOnlyGpuSaveTensorFp16PreprocessImage)
             const std::wstring modelPath = CURRENT_PATH + L"SqueezeNet_fp16.onnx";
             const std::wstring inputPath = CURRENT_PATH + L"fish.png";
+            const std::wstring tensorDataPath = TENSOR_DATA_PATH + L"\\" + METHOD_NAME;
             const std::wstring command = BuildCommand({ EXE_PATH, L"-model ", modelPath, L"-input", inputPath,
-                                                        L"-SaveTensorData", L"First", TENSOR_DATA_PATH, L"-CPU" });
+                                                        L"-SaveTensorData", L"First", tensorDataPath, L"-GPU",
+                                                        L"-inputImagePreprocess 1 0 0 0", L"-BGR" });
             Assert::AreEqual(S_OK, RunProc((wchar_t*)command.c_str()));
-            Assert::AreEqual(true, CompareTensorsFP16(L"OutputTensorData\\Squeezenet_fp16_fish_input_CPU.csv",
-                                                      TENSOR_DATA_PATH + L"\\softmaxout_1CpuIteration1.csv"));
+            Assert::AreEqual(true, CompareTensorsFP16(L"OutputTensorData\\Squeezenet_fp16_fish_input_GPU.csv",
+                                                      tensorDataPath + L"\\softmaxout_1GpuIteration1.csv"));
+        }
+        TEST_METHOD_WITH_NAME(ProvidedImageInputOnlyCpuSaveTensorFp16)
+            const std::wstring modelPath = CURRENT_PATH + L"SqueezeNet_fp16.onnx";
+            const std::wstring inputPath = CURRENT_PATH + L"fish.png";
+            const std::wstring tensorDataPath = TENSOR_DATA_PATH + L"\\" + METHOD_NAME;
+            const std::wstring command = BuildCommand({ EXE_PATH, L"-model ", modelPath, L"-input", inputPath,
+                                                        L"-SaveTensorData", L"First", tensorDataPath, L"-CPU" });
+            Assert::AreEqual(S_OK, RunProc((wchar_t*)command.c_str()));
+            Assert::AreEqual(true, CompareTensors(L"OutputTensorData\\Squeezenet_fp16_fish_input_CPU.csv",
+                                                  tensorDataPath + L"\\softmaxout_1CpuIteration1.csv"));
+        }
+        TEST_METHOD_WITH_NAME(ProvidedImageInputOnlyCpuSaveTensorFp16PreprocessImage)
+            const std::wstring modelPath = CURRENT_PATH + L"SqueezeNet_fp16.onnx";
+            const std::wstring inputPath = CURRENT_PATH + L"fish.png";
+            const std::wstring tensorDataPath = TENSOR_DATA_PATH + L"\\" + METHOD_NAME;
+            const std::wstring command = BuildCommand({ EXE_PATH, L"-model ", modelPath, L"-input", inputPath,
+                                                        L"-SaveTensorData", L"First", tensorDataPath, L"-CPU",
+                                                        L"-inputImagePreprocess 1 0 0 0", L"-BGR" });
+            Assert::AreEqual(S_OK, RunProc((wchar_t*)command.c_str()));
+            Assert::AreEqual(true, CompareTensors(L"OutputTensorData\\Squeezenet_fp16_fish_input_CPU.csv",
+                                                  tensorDataPath + L"\\softmaxout_1CpuIteration1.csv"));
         }
     };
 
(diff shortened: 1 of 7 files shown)


pmbrown1055:
Initially implemented as a command-line argument to WinMLRunner, so that it can be used with existing opset 7 and opset 8 ONNX models.  Please review / discuss.

Tools/WinMLRunner/BindingUtilities.h
```diff
@@ -144,6 +144,35 @@ namespace BindingUtilities
...
+            (meanStdDev[0] != 0 || meanStdDev[1] != 0 || meanStdDev[2] != 0))
+        {
+            Buffer sbBuffer(imgHeight * imgWidth * 4);
+            softwareBitmap.CopyToBuffer(sbBuffer);
+            uint8_t *sbBufferData = sbBuffer.data();
+
+            //Roll the array correctly for the tensor
+            for (int i = 0, count = 0; i < imgHeight * imgWidth; ++i, count += 4)
+            {
+                bindingData[i] = (sbBufferData[count] - meanStdDev[0]) / scale;
```
zhangxiang1993: Can you do an explicit cast here?  
(T)(sbBufferData[count] - meanStdDev[0]) / scale;

Tools/WinMLRunner/BindingUtilities.h
```diff
@@ -144,6 +144,35 @@ namespace BindingUtilities
...
+        T* bindingData = binding.GetData();
+        if (scale != 1.0f &&
+            (meanStdDev[0] != 0 || meanStdDev[1] != 0 || meanStdDev[2] != 0))
+        {
+            Buffer sbBuffer(imgHeight * imgWidth * 4);
+            softwareBitmap.CopyToBuffer(sbBuffer);
+            uint8_t *sbBufferData = sbBuffer.data();
+
+            //Roll the array correctly for the tensor
+            for (int i = 0, count = 0; i < imgHeight * imgWidth; ++i, count += 4)
```
zhangxiang1993: nit: use uint32_t.

ryanlai2:
Hey Paul! Can we have a test for your change? Your earlier comment said that this change gives correct outputs for CV models like DenseNet121, Resnet50. How about Squeezenet? How can we see the range that the model expects? (0,255)? (-1,1)?

Tools/WinMLRunner/CommandLineArgs.cpp
```diff
@@ -30,14 +30,15 @@ void CommandLineArgs::PrintUsage() {
...
+    std::cout << "  -inputImagePreprocess <scale> <msdR> <msdG> <msdB>: float scale factor and per channel meanStdDev factors to preprocess input images before being passed ot the model" << std::endl;
 }
 
 CommandLineArgs::CommandLineArgs()
 {
     int numArgs = 0;
     LPWSTR* args = CommandLineToArgvW(GetCommandLineW(), &numArgs);
 
-    for (int i = 0; i < numArgs; i++)
+    for (int i = 1; i < numArgs; i++)
```
zhangxiang1993: nit: uint32_t

Tools/WinMLRunner/BindingUtilities.h
```diff
@@ -144,6 +144,35 @@ namespace BindingUtilities
...
+        const auto imgWidth = softwareBitmap.PixelWidth();
+        const auto resultSize = imgHeight * imgWidth * 3;
+        if (binding.GetDataBufferSize() != resultSize)
+        {
+            throw hresult_invalid_argument(L"Image Input size/shape is different from what model expects");
+        }
+
+        T* bindingData = binding.GetData();
+        if (scale != 1.0f &&
+            (meanStdDev[0] != 0 || meanStdDev[1] != 0 || meanStdDev[2] != 0))
```
zhangxiang1993: Is it impossible that the actual meanStdDev is {0, 0, 0}? why is scale != 1.0f not sufficient?

Tools/WinMLRunner/BindingUtilities.h
zhangxiang1993: For inputDataType imageBGR, the order of meanStdDev doesn't match.

Tools/WinMLRunner/BindingUtilities.h
pmbrown1055: Yes, I will do this with my next commit.

Tools/WinMLRunner/BindingUtilities.h
pmbrown1055: Ok.  I will do this with my next commit

Tools/WinMLRunner/CommandLineArgs.cpp
pmbrown1055: Ok.

pmbrown1055:
Would need to verify the output result is correct.  SqueezeNet actually requires 0-255 input, so that is operating correctly, and scaling the input will actually cause incorrect results.  I was trying to determine a good way to verify that the predicted result is correct, but am not seeing a great way to do it in the currrent test code (just checking for error messages).  Welcoming any suggestions here.

Since the only models that I've immediately found are DenseNet and ResNet50, should one of them be added to the repo so they can be used to test as well?  DenseNet is a little smaller, but ResNet50 seems to be much more relevant.

pmbrown1055:
One more thing... going back to Paul's comment earlier... this information really should be embedded in the model in the long term.  If an Opset 7 model could contain the scale and meadStdDev information, that would be ideal.  This would only be a workaround for models that don't have the info available on hand.

ryanlai2:
>           Would need to verify the output result is correct.  SqueezeNet actually requires 0-255 input, so that is operating correctly, and scaling the input will actually cause incorrect results.  I was trying to determine a good way to verify that the predicted result is correct, but am not seeing a great way to do it in the currrent test code (just checking for error messages).  Welcoming any suggestions here.
> Since the only models that I've immediately found are DenseNet and ResNet50, should one of them be added to the repo so they can be used to test as well?  DenseNet is a little smaller, but ResNet50 seems to be much more relevant.

Hey Paul, how are you able to determine what kind of inputs that the model expects? Does Netron give you this information? Do you get this information from documentation about the model?

Tools/WinMLRunner/CommandLineArgs.h
```diff
@@ -90,4 +95,7 @@ class CommandLineArgs
     std::wstring m_inputData;
     std::wstring m_outputPath;
     uint32_t m_numIterations = 1;
+
+    float m_scaleFactor = 1.0f;
+    std::array<float, 3> m_meanStdDev{ 0, 0, 0 };
```
smk2007: Please rename or add a comment to indicate that m_meanStdDev corresponds to RBG, BGR, or whatever pixel format.

Alternatively a struct may be more informative here.

smk2007: m_meanStdDev corresponds to 3 floats, are these means or stddevs?? This name is quite confusing.

Tools/WinMLRunner/README.md
```diff
@@ -41,6 +41,7 @@ Required command-Line arguments:
 -silent                  : Silent mode (only errors will be printed to the console)
 -debug                   : Will start a trace logging session.
 -autoScale <mode>        : Will automatically scale an input image to match the required input dimensions of the model.  Pass in the interpolation mode, one of ["Nearest", "Linear", "Cubic", "Fant"].
+-inputImagePreprocess <scale> <msdR> <msdG> <msdB> : Will automatically preprocess the input image before it's sent to the model.  This will enable correct output for models that require this preprocessing.
```
smk2007: Please add more information here as to how this mapping his happening. It is not clear how scale, msdR(G)(B) are being used.

Tools/WinMLRunner/CommandLineArgs.h
```diff
@@ -82,6 +86,7 @@ class CommandLineArgs
     bool m_silent = false;
     bool m_autoScale = false;
     BitmapInterpolationMode m_autoScaleInterpMode = BitmapInterpolationMode::Cubic;
+    bool m_inputImagePreProcess = false;
```
smk2007: nit: i would prefer if the below members are part of a struct, and stored in an std::optional<>. That way you dont need additional tracking boolean flags to see if its been set.

Tools/WinMLRunner/CommandLineArgs.h
```diff
@@ -17,14 +17,18 @@ class CommandLineArgs
...
-   
+    bool InputImagePreProcess() const { return m_inputImagePreProcess; }
+
     const std::wstring& ImagePath() const { return m_imagePath; }
     const std::wstring& CsvPath() const { return m_csvData; }
     const std::wstring& OutputPath() const { return m_outputPath; }
     const std::wstring& FolderPath() const { return m_modelFolderPath; }
     const std::wstring& ModelPath() const { return m_modelPath; }
 
+    BitmapInterpolationMode AutoScaleInterpMode() const { return m_autoScaleInterpMode; }
```
smk2007: nit: AutoScaleInterpolationMode (dont like "Interp")

Tools/WinMLRunner/BindingUtilities.h
```diff
@@ -144,6 +144,35 @@ namespace BindingUtilities
...
+        {
+            Buffer sbBuffer(imgHeight * imgWidth * 4);
```
smk2007: This code is doing 3 copies into 4 buffers. SoftwareBitmap->Buffer->ModelBinding<T>->Tensor*.

The first copy can be eliminated if you use LockBuffer on SoftwareBitmap to access the underlying data.
The last copy can be eliminated if you use ITensorNative to write the transformed data directly into the Tensor* backing memory.

Tools/WinMLRunner/BindingUtilities.h
```diff
@@ -315,6 +343,112 @@ namespace BindingUtilities
...
     }
 
+    ITensor CreateBindableTensorFromImage(
+        const ILearningModelFeatureDescriptor& description,
+        InputDataType inputDataType,
+        const CommandLineArgs args
+    )
+    {
+        auto name = description.Name();
+        const TensorFeatureDescriptor& tensorDescription = description.try_as<TensorFeatureDescriptor>();
```
smk2007: probably shouldnt be a reference here.
should be tensorDescriptor instead of tensorDescription.

auto tensorDescriptor = description.try_as<TensorFeatureDescriptor>();

Tools/WinMLRunner/BindingUtilities.h
```diff
@@ -144,6 +144,35 @@ namespace BindingUtilities
...
 
+    template <typename T>
+    void PreProcessImageToBinding(const SoftwareBitmap& softwareBitmap, ModelBinding<T>& binding, float scale, const std::array<float, 3>& meanStdDev)
+    {
+        const auto imgHeight = softwareBitmap.PixelHeight();
+        const auto imgWidth = softwareBitmap.PixelWidth();
+        const auto resultSize = imgHeight * imgWidth * 3;
+        if (binding.GetDataBufferSize() != resultSize)
+        {
+            throw hresult_invalid_argument(L"Image Input size/shape is different from what model expects");
```
smk2007: This will result in an unhandled exception here since this code path is not part of a try block.

IMHO it would be okay to throw and the calling code should try/catch these and report errors. Cases that cause unexpected program termination make sense to be handled with exceptions (and their exception messages should be reported to the user).

Additionally it would mean that other parts of the code like: 

        const TensorFeatureDescriptor& tensorDescription = description.try_as<TensorFeatureDescriptor>();

        if (!tensorDescription)
        {
            std::cout << "BindingUtilities: Input Description type isn't tensor." << std::endl;
            throw;
        }

Should directly throw and not write to the output directly (but instead rely on app level try/catch blocks to report errors to the user.

Tools/WinMLRunner/BindingUtilities.h
```diff
@@ -315,6 +343,112 @@ namespace BindingUtilities
         return ImageFeatureValue::CreateFromVideoFrame(videoFrame);
     }
 
+    ITensor CreateBindableTensorFromImage(
+        const ILearningModelFeatureDescriptor& description,
+        InputDataType inputDataType,
+        const CommandLineArgs args
```
smk2007: This is a sizeable object. const CommandLineArgs& would be better.

Tools/WinMLRunner/CommandLineArgs.h
```diff
@@ -17,14 +17,18 @@ class CommandLineArgs
...
+    float Scale() const { return m_scaleFactor; }
+    const std::array<float, 3>& MeanStdDev() const { return m_meanStdDev; }
+
```
smk2007: nit: pls remove whitespace

Tools/WinMLRunner/BindingUtilities.h
pmbrown1055: Oops, missed this one.  Will definitely fix it.

Tools/WinMLRunner/CommandLineArgs.h
pmbrown1055: ok

Tools/WinMLRunner/BindingUtilities.h
pmbrown1055: Amusingly I changed this to tensorDescription from tensorDescriptor to better match the description argument, which was a copy from another function in the code.  I'll happily change it back.

Tools/WinMLRunner/CommandLineArgs.h
pmbrown1055: ok

Tools/WinMLRunner/CommandLineArgs.h
pmbrown1055: These are actually precalculated factors that are accounting for the training dataset.  Here's a link to an example post from pytorch on how to pre-interpret the data.  I've seen that actual math that takes the means and standard deviation of the training data set to get these numbers, but haven't located it, yet.

Open to any suggestions here.

https://discuss.pytorch.org/t/how-to-preprocess-input-for-pre-trained-networks/683

pmbrown1055:
I've also known unit tests that were able to snoop specific values in buffers / memory... will this test environment support that?  Or is WinMLRunner dependent on the return value of the exe?

Tools/WinMLRunner/BindingUtilities.h
zhangxiang1993: You can take a look at the code in function `SoftwareBitmapToSoftwareTensor()` at [here](https://github.com/Microsoft/Windows-Machine-Learning/blob/master/Samples/CustomTensorization/CustomTensorization/TensorConvertor.cpp)

Tools/WinMLRunner/BindingUtilities.h
zhangxiang1993: Yes, you can reorder them to RGB. Just make sure they are consistent. And you don't need to worry about the pixelFormat the model wants because Winml will handle that.

zhangxiang1993:
> Would need to verify the output result is correct. SqueezeNet actually requires 0-255 input, so that is operating correctly, and scaling the input will actually cause incorrect results. I was trying to determine a good way to verify that the predicted result is correct, but am not seeing a great way to do it in the currrent test code (just checking for error messages). Welcoming any suggestions here.
> 
> Since the only models that I've immediately found are DenseNet and ResNet50, should one of them be added to the repo so they can be used to test as well? DenseNet is a little smaller, but ResNet50 seems to be much more relevant.

For checking the output result. You can compare the output SoftwareBitmap with the expected SoftwareBitmap(created from ground truth image) byte by byte. And if the error rate is low enough, you can say the output result is correct.

Tools/WinMLRunner/BindingUtilities.h
pmbrown1055: My understanding is that WinML would only reorder the inputs if the input data was the VideoFrame datatype.  These are being passed to WinML as a tensor, so I'm assuming that I will need to reorder the channels if it's required by the model in this code.

Tools/WinMLRunner/BindingUtilities.h
pmbrown1055: Great, thanks for the pointer!  In trying to keep the code as simple as possible, does a VectorView of the Tensor create a copy?  I'm asking because the conversion to float16 using the byte interface will require a specialized template function with a manual conversion.  I would rather keep this in the VectorView<float> to Tensor<Float16> code (I'm assuming it will happen automatically)?

pmbrown1055:
Updated per all suggested changes, and has passing unit tests.  Ready for code review / next steps.

Tools/WinMLRunner/src/CommandLineArgs.cpp
```diff
@@ -69,7 +70,7 @@ void CheckAPICall(int return_value)
 
 CommandLineArgs::CommandLineArgs(const std::vector<std::wstring>& args)
 {
-    for (UINT i = 0; i < args.size(); i++)
```
ryanlai2: Since we don't expect i<0 to be used. Would it be okay to change this type to uint32_t?

Tools/WinMLRunner/src/CommandLineArgs.cpp
```diff
@@ -255,7 +257,14 @@ CommandLineArgs::CommandLineArgs(const std::vector<std::wstring>& args)
             std::wcout << L"Version: " << pFileVersion << "." << pProductVersion << std::endl;
 
             delete[] pVersionData;
-            return;
+        }
+        else if ((_wcsicmp(args[i].c_str(), L"-InputImagePreprocess") == 0) && (i + 4 < args.size()))
+        {
+            m_inputImagePreprocess = true;
```
ryanlai2: CheckNextArgument(args, <index position>) can be used to check that the next 4 arguments aren't flags. Maybe extra checking will be needed to check if the next 4 arguments are numerical.

CheckNextArgument(args, i);
CheckNextArgument(args, i +1);
CheckNextArgument(args, i +2);
CheckNextArgument(args, i +3);

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
```diff
@@ -6,6 +6,7 @@
 #include <Winbase.h>
```
ryanlai2: Would be great to have PreprocessImage tests for DenseNet121 as well! The onnx zoo model is ~32MB large so it's under Github's file limit of 100MB

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
```diff
@@ -747,45 +801,45 @@ namespace WinMLRunnerTest
             const std::wstring command = BuildCommand({ EXE_PATH, L"-model", modelPath, L"-input", inputPath });
             Assert::AreEqual(HRESULT_FROM_WIN32(ERROR_INVALID_PARAMETER), RunProc((wchar_t *)command.c_str()));
         }
-        TEST_METHOD(ProvidedCSVInputGPUSaveTensor)
-        {
+        TEST_METHOD_WITH_NAME(ProvidedCSVInputSaveTensor)
```
ryanlai2: The test name keep GPU in its name so that it is easy to tell that the test runs WinMLRunner using GPU as device.

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
```diff
@@ -235,6 +239,13 @@ namespace WinMLRunnerTest
             std::string copyCommand = "rd /s /q ";
             copyCommand += std::string(INPUT_FOLDER_PATH.begin(), INPUT_FOLDER_PATH.end());
             system(copyCommand.c_str());
+            try
```
ryanlai2: There is already code to remove tensor output files on line 259. It would be redundant to have both class and method clean up functionality for tensor output files.

Class clean up would be more efficient but it doesn't catch the case if the user runs only one test. Thoughts?

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
```diff
@@ -747,45 +801,45 @@ namespace WinMLRunnerTest
...
             const std::wstring inputPath = CURRENT_PATH + L"fish.csv";
+            const std::wstring tensorDataPath = TENSOR_DATA_PATH + L"\\" + METHOD_NAME;
             const std::wstring command = BuildCommand({ EXE_PATH, L"-model", modelPath, L"-input", inputPath,
-                                                        L"-SaveTensorData", L"First", TENSOR_DATA_PATH, L"-GPU" });
-            Assert::AreEqual(S_OK, RunProc((wchar_t*)command.c_str()));
-            Assert::AreEqual(true, CompareTensors(L"OutputTensorData\\Squeezenet_fish_input_GPU.csv",
-                                                  TENSOR_DATA_PATH + L"\\softmaxout_1GpuIteration1.csv"));
+                                                        L"-SaveTensorData", L"First", tensorDataPath, L"-GPU" });
+            Assert::AreEqual(S_OK, RunProc((wchar_t *)command.c_str()));
+            Assert::AreEqual(true, CompareTensors(L"softmaxout1GpuIteration1.csv",
```
ryanlai2: Output file should still remain OutputTensorData\\Squeezenet_fish_input_GPU.csv

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
```diff
@@ -6,6 +6,7 @@
 #include <Winbase.h>
 #include <filesystem>
 #include <fstream>
+#include <iostream>
```
ryanlai2: I don't think iostream is being used.

Tools/WinMLRunner/src/BindingUtilities.h
```diff
@@ -3,69 +3,132 @@
 #include <time.h>
 #include "Common.h"
 #include "ModelBinding.h"
+#include <MemoryBuffer.h>
 #include "Windows.AI.Machinelearning.Native.h"
+#include "DirectXPackedVector.h"
```
ryanlai2: "DirectXPackedVector.h" already included inside Common.h

ryanlai2: Since BindingUtilities is a header file and run.cpp already includes BindingUtilities.h as well as (Common.h, OutputHelper.h, ModelBinding.h) and when the compiler goes and compiles run.cpp, these header files are already included. No need to include OutputHelper.h and CommandLineArgs here.

Tools/WinMLRunner/src/BindingUtilities.h
```diff
@@ -367,11 +542,17 @@ namespace BindingUtilities
...
+        ITensor tensor;
+        SoftwareBitmap softwareBitmap(nullptr);
         std::vector<std::string> elementStrings;
         if (!args.CsvPath().empty())
         {
             elementStrings = ParseCSVElementStrings(args.CsvPath());
         }
+        else if (!args.ImagePath().empty())
+        {
+            softwareBitmap = LoadImageFile(tensorDescriptor, inputDataType, args.ImagePath().c_str(), args, iterationNum);
```
ryanlai2: If inputDataType is Tensor then LoadImageFile will fail. There needs a check in CommandLineArgs to make sure that -inputImagePreprocess flag doesn't get passed in with -Tensor.

Tools/WinMLRunner/src/CommandLineArgs.cpp
pmbrown1055: args.size() returns a size_t (which is signed).  Kept the variable the same to prevent the warning.  I could also call it size_t to explicitly match the output type of args.size(), or cast it if you would prefer that.

Tools/WinMLRunner/src/CommandLineArgs.cpp
pmbrown1055: Do you want me to add the check to see if it is numerical?  I can do that... but I don't think it would be very much more the wtof or sscanf, which would double the string interpretation...

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
pmbrown1055: I can add this if you like... or do you want to use a specific version of DenseNet?  I have opset 7 and opset 8 in fp32 and fp16 readily available (from onnx model zoo).

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
pmbrown1055: May have been a merge error.  Fixed.

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
pmbrown1055: Confirmed that class only cleanup only runs when you run all tests.  I actually think this is better... been very frurstrated trying to debug, while single test runs (failed ones specifically), get deleted.  You don't _have_ to delete like you do without output.csv, since the folder names are unique per test.  I think it would be best to only cleanup at the class level.

Eventually, might want to improve the rest of the tests to leverage the same directory structure capability, so output.csv can exist and the tests can be run in parallel (I don't want to own that change with this change though... maybe another issue)?

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
pmbrown1055: Definitely some kind of merge or intermediate issue.  Fixed.

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
pmbrown1055: Left over from a previous iteration... removed.

Tools/WinMLRunner/src/BindingUtilities.h
pmbrown1055: Removed.

ryanlai2: Eventually refactoring will be needed so that implementation is separated from declaration in .cpp and .h files to avoid confusion like this.

Tools/WinMLRunner/src/BindingUtilities.h
pmbrown1055: -Tensor itself will only work with an input of csv or garbage data, right?  Images are not valid with that option?  Interestingly enough, the way we're testing -inputImagePreprocess right now with 1 0 0 0, is actually generating a tensor and passing that to WinML… is that a capability that can be leveraged to basically do OnClient Tensorization?  Is this interesting at all?

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
ryanlai2: Thanks for the change to make folder names unique per test! However, I'm running into the issue where if I run the same test twice without cleanup, the previous run(s) append to the same output tensor file and contaminates the test result. Thoughts on how to address this?

Tools/WinMLRunner/src/BindingUtilities.h
pmbrown1055: I think that's a good idea.  Forward declaring templates is actually a great way to do this, and to cut down on unnecessary compilation of templates that are not kept (template functions are actually generated every time the header is included in a cpp file, then the linker discards all but one).  Anyway... a little refactoring may be a very good thing. :)

Tools/WinMLRunner/src/CommandLineArgs.cpp
ryanlai2: Seems like according to this: https://stackoverflow.com/questions/409348/iteration-over-stdvector-unsigned-vs-signed-index-variable

std::vector<std::wstring>::size_type is the appropriate type for the index variable? thoughts? The function signature of CheckNextArgument may need to be changed to relect this.

Tools/WinMLRunner/src/CommandLineArgs.cpp
ryanlai2: I would like some additional checking that the next 4 arguments are not flags.

As for the check to see if it's numerical, according to this: https://stackoverflow.com/questions/4654636/how-to-determine-if-a-string-is-a-number-with-c Seems like the most efficient way is to loop through the wstring until a non-digit character is found. [iswdigit](http://www.cplusplus.com/reference/cwctype/iswdigit/) could be used in a helper function to accomplish this. Thoughts?

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
ryanlai2: I think for testing that image preprocessing works, I think using DenseNet ai.onnx version 7 in fp32 would be sufficient. Do you agree?

Tools/WinMLRunner/src/BindingUtilities.h
ryanlai2: Good point! -Tensor itself will only work with an input of csv or garbage data. Adding a check to make sure that -Tensor only works with csv or garbage data inputs would protect -InputImagePreprocess.

Are you proposing to rename -InputImagePreprocess to something like -OnClientTensorize ?

Tools/WinMLRunner/src/CommandLineArgs.cpp
pmbrown1055: I think that making that change is fine.  Will proceed.

Tools/WinMLRunner/src/CommandLineArgs.cpp
pmbrown1055: Already added CheckNextArugment checks to the code (in my local copy).  iwdigit doesn't handle negative or decimals... can search string for not of "0123456789.-", but that negative number matches flag for arugments... lol

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
pmbrown1055: I think that will be fine.  I will go ahead and add a model and test for this.

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
pmbrown1055: Would need to add a CleanupMethod to delete the specific test data before executing... or re-add the Cleanup per method, which would only delete the single test directory... hmmm

Tools/WinMLRunner/src/BindingUtilities.h
pmbrown1055: I like it... no problem changing to OnClientTensorize.  Is your intention that we would implement both, or just the one?

Tools/WinMLRunner/src/CommandLineArgs.cpp
ryanlai2: Does your feature's design suppoed to support negative numbers for the r/g/b factors?

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
ryanlai2: Thanks!

Tools/WinMLRunner/src/BindingUtilities.h
ryanlai2: The usage is inconsistent with the README:  \<msdR\> \<msdG\> \<msdB\> is  \<rFactor\> \<gFactor\> \<bFactor\>.

My intention is to keep your implementation but to just make \<scale\> \<rFactor\> \<gFactor\> \<bFactor\> optional parameters with -OnClientTensorize.

For example, if the user just specifies -OnClientTensorize, then scale = 1, rFactor = 0, gFactor = 0, bFactor = 0

ryanlai2:
Closing because this PR is now stale. https://github.com/microsoft/Windows-Machine-Learning/pull/223 is used to track the on client tensorize change.

PR #97, opened 2018-11-25
User/xianz/winmldb run

zhangxiang1993:


Tools/WinMLDashboard/src/view/edit/LeftPanel.tsx
```diff
@@ -70,32 +72,23 @@ class LeftPanel extends React.Component<IComponentProperties, {}> {
...
+        const name = 'Input And Output';
 
+        const input = this.props.modelInputs;
+        const output = this.props.modelOutputs;
         const inputsForm = this.buildConnectionList(input);
         const outputsForm = this.buildConnectionList(output);
         return (
             <div>
-                <Label className='PanelName'>{name}</Label>
+                <Label className='PanelName' onClick={this.toggleLeft}>{name}</Label>
```
kumraj: This is not be needed if "Edit" button toggles between "View Only" mode and "Edit" mode. View Only mode gives the graph browsing experience. Edit mode should allow both editing and browsing (browsing without Netron's flyout).

The "Edit" button (next to zoom buttons) should be the only place to toggle from Edit mode to View only mode. Currently, just touching any node in the graph switches edit mode to view only mode which makes using the edit mode a bit cumbersome (as mere browsing the graph can lead to switching the mode inadvertently).

PR #102, opened 2018-12-05
Customop sample

orilevari:
Custom operator sample which registers multiple custom operators into the same provider which is used to initialize the learning model. Note that the operator versions of the two new operators (noisyrelu and debug) different so that we do not register the same operator set schema twice. We may want in the future to look into a better mechanism of registering multiple custom operators into the same operator schema sets in separate api calls.


Samples/custom-operator-cpu-sample/desktop/cpp/README.md
```diff
@@ -0,0 +1,52 @@
+# SqueezeNet Object Detection sample
+
+This is a desktop application that uses models containing custom operator definitions and implementations.
+
+To reuse the debug operator you will need to compile debug_cpu.cpp and debug_cpu.h into your project.
```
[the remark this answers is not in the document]
orilevari: Is this clear enough?

Samples/custom-operator-cpu-sample/customize_model/debug_all_outputs.py
```diff
@@ -0,0 +1,40 @@
+# this script will interlace debug operators for every intermediate output of the onnx graph
+
+import sys
+import os.path as path
+import onnx
+from onnx import helper
+import argparse
+
```
kumraj: Remove extra line here.

Samples/custom-operator-cpu-sample/customize_model/debug_all_outputs.py
```diff
@@ -0,0 +1,40 @@
...
+import sys
+import os.path as path
+import onnx
+from onnx import helper
+import argparse
+
+
+def create_modified_model(args):
+
```
kumraj: to be consistent with rest of the code, remove extra line from here.

Samples/custom-operator-cpu-sample/customize_model/debug_all_outputs.py
```diff
@@ -0,0 +1,37 @@
...
+    # gather each unique intermediate output in the graph
+    for node in model.graph.node:
+        for output in node.output:
+            intermediate_outputs.add(output)
+
+    # create a debug operator that consumes each intermediate output
+    # debug operator file path attribute is constructed by the name of the intermediate output
+    for output in intermediate_outputs:
+        inserted_node = model.graph.node.add()
+        valid_output_path = output.replace('/', '\\')
```
youngkim93: Just curious if this was necessary? From my experience both '\\\\' and '/' worked to get directory path

Samples/custom-operator-cpu-sample/customize_model/debug_all_outputs.py
```diff
@@ -0,0 +1,37 @@
...
+        inserted_node.CopyFrom(helper.make_node('Debug', [output], ['unused_' + output], file_type=args.debug_file_type, file_path=valid_output_path + '\\' + valid_output_path))
```
youngkim93: Can we use path.join as we did in the lines above or was this necessary?

Samples/custom-operator-cpu-sample/customize_model/debug_all_outputs.py
orilevari: When I used this script on SqueezeNet which have many outputs which use '/' the file IO failed when running the model with the debug operator provider.

PR #106, opened 2018-12-07
open file with WinML Dashboard

zhangxiang1993:


zhangxiang1993:
image

In the UI part, I added ONNX Version and Quantization options.

The purpose is that we want to support to convert to both ONNX 1.2 and ONNX1.3 no matther what OS build users have(RS5 or 19H1), though users might not be able to run the converted model. The same as quantization, we want to support to convert model with dequantize op.

PR #107, opened 2018-12-10
User/orilevari/cpu fallback enum devices

orilevari:
Introduce an automatic fallback option to CPU device if no GPUs are available. Also remove new lines and include labels.txt in build copy.


Samples/AdapterSelection/AdapterSelection/cpp/main.cpp
```diff
@@ -23,10 +20,10 @@ void LoadLabels();
 VideoFrame LoadImageFile(hstring filePath);
 void PrintResults(IVectorView<float> results);
 bool ParseArgs(int argc, char* argv[]);
+LearningModelDevice getLMDFromAdapterIndex(int index, IDXGIFactory *pFactory);
```
martinb35: please spell out "LearningModelDevice"

Samples/AdapterSelection/AdapterSelection/cpp/main.cpp
```diff
@@ -51,42 +47,24 @@ int main(int argc, char* argv[])
...
-	queueDesc.Type = D3D12_COMMAND_LIST_TYPE_DIRECT;
-	com_ptr<ID3D12CommandQueue> spCommandQueue;
-	spD3D12Device->CreateCommandQueue(&queueDesc, IID_PPV_ARGS(spCommandQueue.put()));
-
-	// create LearningModelDevice from command queue	
-	com_ptr<ILearningModelDeviceFactoryNative> dFactory =
-		get_activation_factory<LearningModelDevice, ILearningModelDeviceFactoryNative>();
-	com_ptr<::IUnknown> spLearningDevice;
-	dFactory->CreateFromD3D12CommandQueue(spCommandQueue.get(), spLearningDevice.put());
+		printf("There are no available adapters, running on CPU...\n");
```
martinb35: how did you test this? do you have a device with no GPU?

Samples/AdapterSelection/AdapterSelection/cpp/main.cpp
```diff
@@ -124,6 +102,31 @@ int main(int argc, char* argv[])
...
 
+LearningModelDevice getLearningModelDeviceFromAdapterIndex(int index, IDXGIFactory *pFactory) {
+
+    // create D3D12Device
+    com_ptr<IDXGIAdapter> spAdapter;
+    pFactory->EnumAdapters(index, spAdapter.put());
+    com_ptr<IUnknown> spIUnknownAdapter;
+    spAdapter->QueryInterface(IID_IUnknown, spIUnknownAdapter.put_void());
+    com_ptr<ID3D12Device> spD3D12Device;
+    D3D12CreateDevice(spIUnknownAdapter.get(), D3D_FEATURE_LEVEL_12_0, _uuidof(ID3D12Device), spD3D12Device.put_void());
```
walrusmcd: >D3D_FEATURE_LEVEL_12_0 [](start = 47, length = 22)

Curious : Why did you choose feature level 12?   Inside the winml runtime we create the device using feature level 11.

[https://docs.microsoft.com/en-us/windows/desktop/direct3d11/overviews-direct3d-11-devices-downlevel-intro](https://docs.microsoft.com/en-us/windows/desktop/direct3d11/overviews-direct3d-11-devices-downlevel-intro)

Samples/AdapterSelection/AdapterSelection/cpp/main.cpp
```diff
@@ -124,6 +102,31 @@ int main(int argc, char* argv[])
...
+    // create D3D12 command queue from device
+    D3D12_COMMAND_QUEUE_DESC queueDesc = {};
+    queueDesc.Flags = D3D12_COMMAND_QUEUE_FLAG_NONE;
+    queueDesc.Type = D3D12_COMMAND_LIST_TYPE_DIRECT;
+    com_ptr<ID3D12CommandQueue> spCommandQueue;
+    spD3D12Device->CreateCommandQueue(&queueDesc, IID_PPV_ARGS(spCommandQueue.put()));
+
+    // create LearningModelDevice from command queue	
+    com_ptr<ILearningModelDeviceFactoryNative> dFactory =
+        get_activation_factory<LearningModelDevice, ILearningModelDeviceFactoryNative>();
```
walrusmcd: >LearningModelDevice [](start = 31, length = 19)

awesome !

Samples/AdapterSelection/AdapterSelection/cpp/main.cpp
walrusmcd: you want to do some more testing on this, since DXGI will most likely return to you the basic render device (warp) .   which we don't use for ML.

[https://docs.microsoft.com/en-us/windows/desktop/api/d3d12/nf-d3d12-d3d12createdevice](https://docs.microsoft.com/en-us/windows/desktop/api/d3d12/nf-d3d12-d3d12createdevice)


---
In reply to: [240405112](https://github.com/Microsoft/Windows-Machine-Learning/pull/107#discussion_r240405112) [](ancestors = 240405112)

Samples/AdapterSelection/AdapterSelection/cpp/main.cpp
orilevari: hmm, I assumed that since its a d3d12 device that I would want feature level 12. This sample uses a D3D12 Command Queue to construct a LearningModelDevice, so I do need it to be a d3d12 device.

Samples/AdapterSelection/AdapterSelection/cpp/main.cpp
orilevari: oh yes, it does return that, and that is the one that runs really slow like I mentioned. Should I not include that in the list of adapters the user can choose?

walrusmcd: right, you want to filter out WARP.  don't offer that as a choice.

---
In reply to: [240839063](https://github.com/Microsoft/Windows-Machine-Learning/pull/107#discussion_r240839063) [](ancestors = 240839063)

Samples/AdapterSelection/AdapterSelection/cpp/main.cpp
walrusmcd: nope, your good with feature level 11.   i would go with that, since it's the lowest common demoninator and gets us the most support for video cards.   feature level 11_0 is the very first feature level that d3d12 supported.   

[https://en.wikipedia.org/wiki/Feature_levels_in_Direct3D](https://en.wikipedia.org/wiki/Feature_levels_in_Direct3D) 

---
In reply to: [240838624](https://github.com/Microsoft/Windows-Machine-Learning/pull/107#discussion_r240838624) [](ancestors = 240838624)

Samples/AdapterSelection/AdapterSelection/cpp/main.cpp
orilevari: https://docs.microsoft.com/en-us/windows/desktop/direct3ddxgi/d3d10-graphics-programming-guide-dxgi
I filter out software drivers using both the adapter description DXGI_ADAPTER_FLAG_SOFTWARE flag and the vendorid and deviceid. I think this should be sufficient to guarantee that all enumerated adapters are valid GPUs.

orilevari: Confirmed that when run on a vm without access to GPUs the software adapter is correctly suppressed and execution will auto fallback to CPU.

PR #108, opened 2018-12-10
Remove redundant Wall Clock Time metrics

ryanlai2:
This PR removes redundant Wall Clock Time metrics because the profiler already collects the time.

### DIFF
--- a/Tools/WinMLRunner/Main.cpp
+++ b/Tools/WinMLRunner/Main.cpp
@@ -337,6 +337,7 @@ HRESULT EvaluateModels(
                     {
                         if (args.PerfCapture())
                         {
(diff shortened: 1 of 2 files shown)


Tools/WinMLRunner/Main.cpp
```diff
@@ -347,8 +337,8 @@ HRESULT EvaluateModels(
                     {
                         if (args.PerfCapture())
                         {
-                            output.ResetBindAndEvalTImes();
-                            g_Profiler.Reset();
+                            //Resets all values from profiler for bind and evaluate.
```
youngkim93: nit: // Resets all values...

ryanlai2: Fixed

PR #117, opened 2019-01-04
User/orilevari/top3 bug

orilevari:
Fixes a bug in our samples that use Squeezenet where the top 3 probabilities that are extracted from the result vector can be inaccurate in many cases due to a bug in the sorting algorithm implemented. When I fixed the bug the output actually was different for the third most probable label when running with the kitten picture.

### DIFF
--- a/Samples/AdapterSelection/AdapterSelection/cpp/README.md
+++ b/Samples/AdapterSelection/AdapterSelection/cpp/README.md
@@ -53,7 +53,7 @@ Note: SqueezeNet was trained to work with image sizes of 224x224, so you must pr
   model run took 31 ticks
   tabby, tabby cat with confidence of 0.931461
   Egyptian cat with confidence of 0.065307
-  Persian cat with confidence of 0.000193
+  tiger cat with confidence of 0.002927
   ```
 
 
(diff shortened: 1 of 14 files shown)


Samples/AdapterSelection/AdapterSelection/cpp/README.md
```diff
@@ -53,7 +53,7 @@ Note: SqueezeNet was trained to work with image sizes of 224x224, so you must pr
   model run took 31 ticks
   tabby, tabby cat with confidence of 0.931461
   Egyptian cat with confidence of 0.065307
-  Persian cat with confidence of 0.000193
+  tiger cat with confidence of 0.002927
```
youngkim93: So the actual top 3 was not displayed correctly :)

Samples/AdapterSelection/AdapterSelection/cpp/main.cpp
```diff
@@ -205,30 +205,31 @@ VideoFrame LoadImageFile(hstring filePath)
...
-	}
-	// Display the result
-	for (int i = 0; i < 3; i++)
-	{
-		printf("%s with confidence of %f\n", labels[topProbabilityLabelIndexes[i]].c_str(), topProbabilities[i]);
-	}
+    // load the labels
+    LoadLabels();
+
+    struct result {
```
youngkim93: nit: you can probably use std::pair instead of struct

PR #119, opened 2019-01-05
Refactor WinMLRunner to build DLL and EXE from static library

ryanlai2:
There were a few changes in this PR:

1.  Changed previous main method from WinMLRunner to be a "run" method. This "run" method takes in a CommandLineArgs object by reference.

2. Created a project called WinMLRunnerStaticLib that builds WinMLRunner's source code (without main method) into a static library.

3. WinMLRunner project still builds WinMLRunner.exe but takes WinMLRunnerStaticLib project as a reference. It also has a main.cpp file that generates a CommandLineArgs object and passes into the "run" method call.

4. Created a project called WinMLRunnerDLL that also takes WinMLRunnerStaticLib as a reference. It has a dllmain.cpp file that is the entry point for this DLL.  __declspec(dllexport) was added to "run" method and "CommandLineArgs" constructor so that these can be exported and exposed. If a project were to use these entry points, the project would need to static link WinMLRunnerDLL.lib. For now, DLLMain just returns True. This can be fleshed out in future iterations.

5. Refactored test code so that the all tests use the "BuildCommand" helper function to construct command. This makes test code cleaner.

6. Moved source code into a folder called "src".

7. Static casted types to resolve build warnings.

8. These changes also build ARM64 version of WinMLRunner.

9. For Tensor CSV inputs, if the size didn't match what the model expects, then the exception wasn't caught. This PR contains a change to fix this.


Tools/WinMLRunner/src/Run.cpp
```diff
@@ -280,7 +280,17 @@ HRESULT EvaluateModel(
...
 
-        std::vector<ILearningModelFeatureValue> inputFeatures = GenerateInputFeatures(model, args, inputBindingType, inputDataType, winrtDevice);
+        std::vector<ILearningModelFeatureValue> inputFeatures;
+        try
+        {
+            inputFeatures = GenerateInputFeatures(model, args, inputBindingType, inputDataType, winrtDevice);
+        }
+        catch (hresult_error hr)
+        {
+            std::cout << "\nGenerating Input Features [FAILED]" << std::endl;
```
youngkim93: Le'ts use wcout here for consistency

Tools/WinMLRunner/src/main.cpp
```diff
@@ -0,0 +1,8 @@
+#include "CommandLineArgs.h"
+#include "Run.h"
+
+int main(int argc, char** argv)
```
youngkim93: Is this a main function for winmlrunner.exe? are we using argv here?

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
```diff
@@ -511,15 +503,13 @@ namespace WinMLRunnerTest
...
-            std::wstring command = curPath + L"./WinMLRunner " + L"-model invalid_model_name";
+            const std::wstring command = BuildCommand({ EXE_PATH, L"-model", L"invalid_model_name" });
             Assert::AreNotEqual(0, RunProc((wchar_t *)command.c_str()));
         }
 
         TEST_METHOD(TestPrintUsage)
         {
-            auto const curPath = FileHelper::GetModulePath();
-            std::wstring command = curPath + L"./WinMLRunner";
+            const std::wstring command = BuildCommand({ EXE_PATH});
```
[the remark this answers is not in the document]
ryanlai2: Fixed!

Tools/WinMLRunner/src/main.cpp
ryanlai2: You're right, this is the main function. Fixed!

Tools/WinMLRunner/src/Run.cpp
ryanlai2: Changed.

Tools/WinMLRunner/WinMLRunner.sln
```diff
@@ -20,8 +25,8 @@ Global
 		Release|x86 = Release|x86
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
-		{81EA9CC6-8A26-4583-B1A4-84740EF815C8}.Debug|ARM64.ActiveCfg = Debug|ARM64
-		{81EA9CC6-8A26-4583-B1A4-84740EF815C8}.Debug|ARM64.Build.0 = Debug|ARM64
+		{81EA9CC6-8A26-4583-B1A4-84740EF815C8}.Debug|ARM64.ActiveCfg = Release|ARM64
+		{81EA9CC6-8A26-4583-B1A4-84740EF815C8}.Debug|ARM64.Build.0 = Release|ARM64
 		{81EA9CC6-8A26-4583-B1A4-84740EF815C8}.Debug|x64.ActiveCfg = Debug|x64
```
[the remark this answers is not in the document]
ryanlai2: Yep! Fixed

PR #120, opened 2019-01-08
Add -terse flag option to suppress repetitive console output

pmbrown1055:
… to the Console.  Default to -normal.

### DIFF
--- a/Tools/WinMLRunner/README.md
+++ b/Tools/WinMLRunner/README.md
@@ -38,9 +38,9 @@ Required command-Line arguments:
 -input <image/CSV path>  : Will bind image/data from CSV to model.
 -output <CSV path>       : Path to the CSV where the perf results will be written.
 -IgnoreFirstRun          : Will ignore the first run in the perf results when calculating the average
--silent                  : Silent mode (only errors will be printed to the console)
 -savePerIterationPerf	 : Save per iteration performance results to csv file.
 -debug                   : Will start a trace logging session.
+-terse                   : Will suppress repetitive console output (initial iteration and summary info will be output).
 -autoScale <mode>        : Will automatically scale an input image to match the required input dimensions of the model.  Pass in the interpolation mode, one of ["Nearest", "Linear", "Cubic", "Fant"].
 
  ```
--- a/Tools/WinMLRunner/src/OutputHelper.h
+++ b/Tools/WinMLRunner/src/OutputHelper.h
@@ -15,9 +15,8 @@ using namespace winrt::Windows::Storage::Streams;
 class OutputHelper
 {
 public:
-    OutputHelper(int numIterations, bool silent)
+    OutputHelper(int numIterations)
     {
-        m_silent = silent;
         m_clockLoadTimes.resize(numIterations, 0.0);
         m_clockBindTimes.resize(numIterations, 0.0);
         m_clockEvalTimes.resize(numIterations, 0.0);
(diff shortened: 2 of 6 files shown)


pmbrown1055:
Intent is to clean up the default console output to just be summary data, maintaining silent mode for no output, and a verbose mode to print all iteration information to the console.

ryanlai2:
We've decided to deprecate the "-silent" flag and we want the verbosity option to be coupled with the "-perf" flag.
By default, All iteration pass/fail information to the console.
 
"-perf" will print performance numbers without all the noise of iteration information, pass/fail
"-perf verbose" will print iteration information with performance numbers
 
How does this sound?

Intent is to keep WinMLRunner by default as a validation tool to check if the provided model / output can be used with WinML. The simplest command (./winmlrunner -model ./Squeezenet) should keep providing enough info to show what passes (load/bind/ eval) and what fails.

pmbrown1055:
No problem with the tuning.  Just to make sure I understand:

Options:
none -- no perf specified on the command line, will not print perf data, but will print out pass / fail information for all iterations.
-perf will print only summary information of the run and leave out the per iteration pass / fail info
-perf verbose will print out all summary and per iteration pass / fail info
-perf silent (?) will only print error messages to the console.

Is the -perf silent mode something that you want to support?  Or will there be no -silent running mode with this change?

ryanlai2:
Options:
- none -- no perf specified on the command line, will not print perf data, but will print out pass / fail information for all iterations. 
   - **Yep this is correct.**

- -perf will print only summary information of the run and leave out the per iteration pass / fail info
    - **Yep! -perf will only print out performance data as well as the load model summary information.**


- -perf verbose will print out all summary and per iteration pass / fail info
   - **Correct, -perf verbose will print out load model summary information, iteration pass / fail info, as well as performance data**


- -perf silent (?) will only print error messages to the console. Is the -perf silent mode something that you want to support?  Or will there be no -silent running mode with this change?
    - **-perf silent model isn't something that we want to support. If it only prints error messages to the console, the message isn't much helpful to the user if there isn't more context of where the error happened (load / session creation /bind / evaluate). Pass/fail information is helpful for this. There will be no -silent running mode after this change.**

Thank you so much Paul!

ryanlai2:
Hi Paul! I have a modification I just wanted to ask your opinion on. :)
Here are the flag options:
- none -- no perf specified on the command line, will not print perf data, but will print out pass / fail information for all iterations.

- -perf will print only summary information and performance info of the run and leave out the per iteration pass / fail info

- -perf **-verbose** will print out all summary and performance info and per iteration pass / fail info. 
   - While verbose flag is only in effect only when the -perf flag is on as well, we can use this -verbose flag to affect other flags as well in the future.

- -silent flag is deprecated.


Thoughts?

pmbrown1055:
Updated with latest discussed changes.  Please take a look.

Tools/WinMLRunner/src/OutputHelper.h
```diff
@@ -30,109 +29,93 @@ class OutputHelper
...
-                "Evaluating (device = %s, iteration = %d, inputBinding = %s, inputDataType = %s, deviceCreationLocation = %s)...",
-                TypeHelper::Stringify(deviceType).c_str(),
-                iteration,
-                TypeHelper::Stringify(inputBindingType).c_str(),
-                TypeHelper::Stringify(inputDataType).c_str(),
-                TypeHelper::Stringify(deviceCreationLocation).c_str()
-            );
-        }
+        printf(
+            "Evaluating (device = %s, iteration = %d, inputBinding = %s, inputDataType = %s, deviceCreationLocation = %s)...%s",
```
ryanlai2:
```suggestion
            "Evaluating (device = %s, iteration = %d, inputBinding = %s, inputDataType = %s, deviceCreationLocation = %s)...%s\n",
```
Need to add \n newline here. If we run without -perf flag, and with multiple configurations, the next configuration's binding console output will be on the same line as the previous evaluate console output

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -295,18 +283,34 @@ HRESULT EvaluateModel(
 
         if (FAILED(bindInputResult))
         {
+            output.PrintBindingInfo(i + 1, deviceType, inputBindingType, inputDataType, deviceCreationLocation, "[FAILED]");
             return bindInputResult;
         }
-
-        output.PrintEvaluatingInfo(i + 1, deviceType, inputBindingType, inputDataType, deviceCreationLocation);
+        else if (i == 0)
```
ryanlai2: Hey @pmbrown1055  would it be possible to keep default behavior for non performance scenarios? By starting to separate here, it would help in the future for refactoring between performance / validation scenarios.
```suggestion
        if (!args.PerfCapture()) //When not capturing performance, we want to keep default behavior.
        {
                output.PrintBindingInfo(i + 1, deviceType, inputBindingType, inputDataType, deviceCreationLocation, "[SUCCESS]"); 
        }
        else if (i == 0) //If capturing performance, we want to decrease amount of printing to console
        {
            output.PrintBindingInfo(i + 1, deviceType, inputBindingType, inputDataType, deviceCreationLocation, "[SUCCESS]");
        }
```

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -295,18 +283,34 @@ HRESULT EvaluateModel(
...
+        }
 
         HRESULT evalResult = EvaluateModel(model, context, session, isGarbageData, args, output, captureIterationPerf, i);
 
         if (FAILED(evalResult))
         {
+            output.PrintEvaluatingInfo(i + 1, deviceType, inputBindingType, inputDataType, deviceCreationLocation, "[FAILED]");
             return evalResult;
         }
+        else if (i == 0)
```
ryanlai2:
```suggestion
        if (!args.PerfCapture()) //When not capturing performance, we want to keep printing out binding information.
        {
            output.PrintEvaluatingInfo(i + 1, deviceType, inputBindingType, inputDataType, deviceCreationLocation, "[SUCCESS]");
        }
        else if (i == 0) //If capturing performance, we want to decrease amount of printing to console
        {
            output.PrintEvaluatingInfo(i + 1, deviceType, inputBindingType, inputDataType, deviceCreationLocation, "[SUCCESS]");
        }
```

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -178,12 +173,7 @@ HRESULT EvaluateModel(
         return hr.code();
     }
 
-    if (!args.Silent())
-    {
-        std::cout << "[SUCCESS]" << std::endl;
-    }
-
-    if (!isGarbageData && !args.Silent())
+    if (capturePerf && !isGarbageData)
```
ryanlai2: If we want to capture performance and we included our own input and then iterated 100 times, processing and printing out the results would be expensive

Tools/WinMLRunner/src/Run.cpp
pmbrown1055: My intention was to clean up the console output for all scenarios... but if you think it's important to keep all the output for all iterations for the default case, then I can leave it in.  If that is the case, then we should probably add some kind of -terse or -silent option back in to all the user to not see the default prints...

ryanlai2: Okay I think that works! How about this for adding the -terse option?:

```suggestion
        if (!args.PerfCapture() && !args.TerseOutput()) //When not capturing performance and the output isn't set to terse, we want to keep printing out binding information.
        {
                output.PrintBindingInfo(i + 1, deviceType, inputBindingType, inputDataType, deviceCreationLocation, "[SUCCESS]"); 
        }
        else if (i == 0) //If capturing performance, we want to decrease amount of printing to console
        {
            output.PrintBindingInfo(i + 1, deviceType, inputBindingType, inputDataType, deviceCreationLocation, "[SUCCESS]");
        }
```

pmbrown1055:
Updated with the latest changes.  Altered the Evaluation flow a little bit to move all output / processing outside of Evaluate.  EvaluationModel inner call now only does perf and evaluate.  Let me know what you think.

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -269,13 +252,14 @@ HRESULT EvaluateModel(
     // Add one more iteration if we ignore the first run
     uint32_t numIterations = args.NumIterations() + args.IgnoreFirstRun();
 
+    bool isGarbageData = args.CsvPath().empty() && args.ImagePath().empty();
```
ryanlai2: In CommandLineArgs, there is a method called IsGarbageInput that we can use now

Tools/WinMLRunner/README.md
```diff
@@ -38,7 +38,6 @@ Required command-Line arguments:
 -input <image/CSV path>  : Will bind image/data from CSV to model.
 -output <CSV path>       : Path to the CSV where the perf results will be written.
 -IgnoreFirstRun          : Will ignore the first run in the perf results when calculating the average
--silent                  : Silent mode (only errors will be printed to the console)
```
ryanlai2: Can we add in the -terse flag with description here?

pmbrown1055:
Updated -terse option to work the same on both default and perf code paths.

PR #121, opened 2019-01-09
Refactor winmlrunner tests to use hresult

ryanlai2:
This PR uses HRESULT to check the status code of Winmlrunner for fail / pass scenarios. This makes it more clear to be able to use Error Constants.


youngkim93:
LGTM

PR #122, opened 2019-01-11
Change so Garbage Tensors don't use extra string stream to generate garbage data

ryanlai2:
Before this PR, WriteDataToBinding was still called even though the intent was to use garbage data. This change makes it so that arbitrary memory is created for the binding. The memory is memset to 0 so that garbage data results are deterministic and consistent.

**There is about a 2x performance increase with evaluate with this change since there is less time between GPU usage. Before this change, the GPU powers down / power up between iterations.**

I ran Squeezenet 300 iterations specifying GPU as device and compared the captured average performance.

Before this PR change:

  Load: 21.470000 ms
  Bind: 0.473307 ms
  **Evaluate: 18.5709 ms**
  Total Time: 40.5142 ms
  Working Set Memory usage (evaluate): 0 MB
  Dedicated Memory Usage (evaluate): 0 MB
  Shared Memory Usage (evaluate): 0 MB

After this PR change:

  Load: 46.911000 ms
  Bind: 0.685665 ms
  **Evaluate: 8.89519 ms**
  Total Time: 56.4919 ms
  Working Set Memory usage (evaluate): 0.00375366 MB
  Dedicated Memory Usage (evaluate): 0 MB
  Shared Memory Usage (evaluate): 0 MB

### DIFF
--- a/Tools/WinMLRunner/src/CommandLineArgs.h
+++ b/Tools/WinMLRunner/src/CommandLineArgs.h
@@ -63,6 +63,12 @@ public:
         return m_createDeviceInWinML || !m_createDeviceOnClient;
     }
 
+    bool IsGarbageInput() const
+    {
+        // When there is no image or csv input provided, then garbage input binding is used.
+        return m_imagePath.empty() && m_csvData.empty();
+    }
+
     uint32_t NumIterations() const { return m_numIterations; }
 
 private:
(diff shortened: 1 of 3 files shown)


Tools/WinMLRunner/src/CommandLineArgs.h
```diff
@@ -63,6 +63,12 @@ class CommandLineArgs
         return m_createDeviceInWinML || !m_createDeviceOnClient;
     }
 
+    bool IsGarbageInput() const
```
smk2007: nit: can we change "garbage" to something else.

PR #125, opened 2019-01-15
User/orilevari/fix build errors

orilevari:
This pull request is just adding one more commit to what was merged into master. I am removing the build step of publishing artifacts after building, since it fails nondeterministically on the emoji sample publishing. I think it would be best if what is checked into master always succeeds, and I can work on fixing this artifact publishing down the road in a different branch.


youngkim93:
A lot of commits here for one change (It seems like this is from the existing branch outdated from master?) I think as long as you squash here you should be fine :)

PR #126, opened 2019-01-15
Before creating session, print out learning model device. If gpu, print out gpu name

ryanlai2:
This change accesses IDirect3DDevice from the LearningModelDevice and tries to print out the GPU name

### DIFF
--- a/Tools/WinMLRunner/src/OutputHelper.h
+++ b/Tools/WinMLRunner/src/OutputHelper.h
@@ -140,7 +140,7 @@ public:
         }
     }
 
-    void PrintLearningModelDevice(DeviceType deviceType, LearningModelDevice& device)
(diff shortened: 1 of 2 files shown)


ryanlai2:
image

image

Tools/WinMLRunner/src/OutputHelper.h
```diff
@@ -136,6 +140,35 @@ class OutputHelper
         }
     }
 
+    void PrintLearningModelDevice(DeviceType deviceType, LearningModelDevice& device)
```
youngkim93: const LearningModelDevice&

ryanlai2: Yep, that makes sense to make it const because the intention isn't to change LearningModelDevice

PR #129, opened 2019-01-18
tidy up label formatting

orilevari:


orilevari:
winmldashboardconvert

PR #134, opened 2019-01-22
Stop automatically outputting performance file for -perf flag

ryanlai2:
-Perf flag would output performance numbers AND automatically create a CSV file. This change separates this so that -perfOutput flag <with optional csv file name> needs to be specified in order for the CSV file to be generated.

More changes contained in the PR:

- If _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS flag was already defined, then don't redefine it.
- "Model Binding" column in the perf csv file should actually be "Device Type"
- For creating d3d device on client and passing into session, this should only be supported if the DeviceType isn't CPU.

### DIFF
--- a/Tools/WinMLRunner/src/CommandLineArgs.cpp
+++ b/Tools/WinMLRunner/src/CommandLineArgs.cpp
@@ -25,7 +25,7 @@ void CommandLineArgs::PrintUsage() {
     std::cout << "  -perf : capture timing measurements" << std::endl;
     std::cout << "  -iterations : # times perf measurements will be run/averaged" << std::endl;
     std::cout << "  -input <fully qualified path>: binds image or CSV to model" << std::endl;
-    std::cout << "  -output <fully qualified path>: csv file to write the perf results to" << std::endl;
+    std::cout << "  -perfOutput optional:<fully qualified path>: csv file to write the perf results to" << std::endl;
     std::cout << "  -IgnoreFirstRun : ignore the first run in the perf results when calculating the average" << std::endl;
     std::cout << "  -savePerIterationPerf : save per iteration performance results to csv file" << std::endl;
     std::cout << "  -debug: print trace logs" << std::endl;
@@ -77,9 +77,13 @@ CommandLineArgs::CommandLineArgs(std::vector<std::wstring>& args)
         {
             m_inputData = args[++i];
         }
-        else if ((_wcsicmp(args[i].c_str(), L"-output") == 0))
+        else if ((_wcsicmp(args[i].c_str(), L"-perfOutput") == 0))
         {
-            m_outputPath = args[++i];
+            if (i + 1 < args.size() && args[++i][0] != *L"-")
+            {
+                m_outputPath = args[++i];
+            }
+            m_perfOutput = true;
         }
         else if ((_wcsicmp(args[i].c_str(), L"-RGB") == 0))
         {
--- a/Tools/WinMLRunner/src/CommandLineArgs.cpp
+++ b/Tools/WinMLRunner/src/CommandLineArgs.cpp
@@ -21,19 +21,19 @@ void CommandLineArgs::PrintUsage() {
     std::cout << "  -GPUBoundInput : bind the input to the GPU" << std::endl;
     std::cout << "  -RGB : load the input as an RGB image" << std::endl;
     std::cout << "  -BGR : load the input as a BGR image" << std::endl;
-    std::cout << "  -tensor : load the input as a tensor" << std::endl;
-    std::cout << "  -perf : capture timing measurements" << std::endl;
-    std::cout << "  -iterations : # times perf measurements will be run/averaged" << std::endl;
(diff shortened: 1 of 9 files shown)


Tools/WinMLRunner/src/CommandLineArgs.cpp
```diff
@@ -77,9 +77,13 @@ CommandLineArgs::CommandLineArgs(std::vector<std::wstring>& args)
         {
             m_inputData = args[++i];
         }
-        else if ((_wcsicmp(args[i].c_str(), L"-output") == 0))
+        else if ((_wcsicmp(args[i].c_str(), L"-perfOutput") == 0))
         {
-            m_outputPath = args[++i];
+            if (i + 1 < args.size() && args[++i][0] != *L"-")
```
youngkim93: Will L'-' work?

Tools/WinMLRunner/src/CommandLineArgs.cpp
```diff
@@ -77,9 +77,13 @@ CommandLineArgs::CommandLineArgs(std::vector<std::wstring>& args)
...
+            {
+                m_outputPath = args[i];
```
youngkim93: m_perfOutputPath?

Tools/WinMLRunner/src/CommandLineArgs.cpp
```diff
@@ -25,7 +25,7 @@ void CommandLineArgs::PrintUsage() {
     std::cout << "  -perf : capture timing measurements" << std::endl;
     std::cout << "  -iterations : # times perf measurements will be run/averaged" << std::endl;
     std::cout << "  -input <fully qualified path>: binds image or CSV to model" << std::endl;
-    std::cout << "  -output <fully qualified path>: csv file to write the perf results to" << std::endl;
+    std::cout << "  -perfOutput optional:<fully qualified path>: csv file to write the perf results to" << std::endl;
```
youngkim93: Let's make camel case consistent.

Tools/WinMLRunner/src/OutputHelper.h
```diff
@@ -377,7 +377,7 @@ class OutputHelper
             if (bNewFile)
             {
                 fout << "Model Name" << ","
-                     << "Model Binding" << ","
+                     << "Device Type" << ","
```
youngkim93: Change parameter name too

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -196,7 +196,8 @@ HRESULT EvaluateModel(
 
     try
     {
-        if (deviceCreationLocation == DeviceCreationLocation::ClientCode)
+        if (deviceCreationLocation == DeviceCreationLocation::ClientCode
+            && (deviceType != DeviceType::CPU))
```
youngkim93: Let's remove parenthesis inside. Just do deviceType != DeviceType::CPU

Tools/WinMLRunner/src/CommandLineArgs.h
```diff
@@ -4,6 +4,7 @@
 class CommandLineArgs
 {
 public:
+    CommandLineArgs() {};
     __declspec(dllexport) CommandLineArgs(std::vector<std::wstring>& args);
```
youngkim93: Let's remove __Declspec(dllexport) below. Also make args const

Tools/WinMLRunner/src/CommandLineArgs.h
```diff
@@ -84,12 +87,17 @@ class CommandLineArgs
...
     void TogglePerIterationPerformanceCapture(const bool perIterCapture) { m_perIterCapture = perIterCapture; }
     void ToggleDebugOutput(const bool debug) { m_debug = debug; }
+    void ToggleTerseOutput(const bool terseOutput) { m_terseOutput = terseOutput; }
 
 
     void SetModelPath(const std::wstring modelPath) { m_modelPath = modelPath; }
     void SetInputDataPath(const std::wstring inputDataPath) { m_inputData = inputDataPath; }
-    void SetPerformanceCSVPath(const std::wstring performanceCSVPath) { m_outputPath = performanceCSVPath; }
-
+    void SetPerformanceCSVPath(const std::wstring performanceCSVPath)
```
youngkim93: const std::wstring &performanceCSVPath

Tools/WinMLRunner/src/CommandLineArgs.h
```diff
@@ -84,12 +87,17 @@ class CommandLineArgs
     void ToggleIgnoreFirstRun(const bool ignoreFirstRun) { m_ignoreFirstRun=ignoreFirstRun;}
     void TogglePerIterationPerformanceCapture(const bool perIterCapture) { m_perIterCapture = perIterCapture; }
     void ToggleDebugOutput(const bool debug) { m_debug = debug; }
+    void ToggleTerseOutput(const bool terseOutput) { m_terseOutput = terseOutput; }
```
youngkim93: bools don't need to be const. Primitive data types are passed as values.

PR #146, opened 2019-01-29
Added versioning info for winmlrunner

ryanlai2:
File and Product Version will be 1.0.0.0 for now. Will change this schema in future iterations. This is so that winmlrunner can be released.

image


ryanlai2:
I queued another build. This one succeeded

PR #159, opened 2019-02-12
Updates to public samples yaml

orilevari:
1) Build artifacts will be published even if one or more samples failed building. This will allow more granularity to the test pipeline.

2) Replaced the template syntax for multi configuration with the built in functionality using the strategy:matrix option.


orilevari:
I also removed the ability to override the TargetPlatformVersion in SqueezeNet.jsproj because the UWP JS props files will override this value to 8.0.

PR #160, opened 2019-02-13
Enhance CreateByClient feature to create a device on the DEFAULT, MIN…

pmbrown1055:
…_POWER, or HIGH_PERFORMANCE adapter in the system when using those command line arguments.

### DIFF
--- a/Tools/WinMLRunner/src/Run.cpp
+++ b/Tools/WinMLRunner/src/Run.cpp
@@ -210,31 +210,41 @@ HRESULT EvaluateModel(
 
         if (deviceCreationLocation == DeviceCreationLocation::ClientCode && deviceType != DeviceType::CPU)
         {
+            // Enumerate Adapters to pick the requested one.
+            com_ptr<IDXGIFactory6> factory;
+            HRESULT hr = CreateDXGIFactory(__uuidof(IDXGIFactory6), factory.put_void());
+            THROW_IF_FAILED(hr);
+
+            com_ptr<IDXGIAdapter> adapter;
+            switch (deviceType)
+            {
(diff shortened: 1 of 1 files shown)


Tools/WinMLRunner/src/Run.cpp
```diff
@@ -210,31 +210,41 @@ HRESULT EvaluateModel(
...
+                throw hresult(E_INVALIDARG);
             }
+            THROW_IF_FAILED(hr);
 
-            com_ptr<IDXGIDevice> dxgiDevice;
-            hr = d3d11Device->QueryInterface(IID_PPV_ARGS(dxgiDevice.put()));
+            // Creating the device on the client and using it to create the video frame and initialize the session makes sure that everything is on
+            // the same device. This usually avoids an expensive cross-device and cross-videoframe copy via the VideoFrame pipeline.
+            com_ptr<ID3D11Device> d3d11Device;
+            hr = D3D11CreateDevice(adapter.get(), D3D_DRIVER_TYPE_UNKNOWN, nullptr, D3D11_CREATE_DEVICE_BGRA_SUPPORT, nullptr, 0, D3D11_SDK_VERSION, d3d11Device.put(), nullptr, nullptr);
```
ryanlai2: Why do we use D3D_DRIVER_TYPE_UNKNOWN enumeration?  We used D3D_DRIVER_TYPE_HARDWARE before this change

pmbrown1055: You can't pass an adapter into the Create call declared as HARDWARE.  It must be declared as unknown for the call to pass.  Verified with D3D team.

PR #166, opened 2019-02-16
User/orilevari/tensorization paths

orilevari:
This modifies the way that the custom tensorization sample resolves it's paths to ingest and output files.

Previously it would rely on its input files (.onnx and .png) to exist in a directory that was resolved in a way that would crash if the executable was not located in a directory which had a parent directory containing a subdirectory with the name CustomTensorization. This was posing a problem for setting up build artifacts to test against.

Now, these input files are copied on build and searched for in the same directory as the executable. Furthermore, the old path resolving mechanism is used for the output files if and only if that parent directory exists, otherwise files will be output in the same directory as the executable.

This preserves the functionality originally intended when the sample is run in the context of this repository, which was to make sure it is easy for users to locate the output files by placing them next to the vcxproj file.

### DIFF
--- a/Samples/CustomTensorization/CustomTensorization/TensorConvertor.cpp
+++ b/Samples/CustomTensorization/CustomTensorization/TensorConvertor.cpp
@@ -20,7 +20,7 @@ EXTERN_C IMAGE_DOS_HEADER __ImageBase;
 
 namespace TensorizationHelper
 {
-    std::wstring GetModulePath(bool visiblePath)
+    std::wstring GetModulePath()
     {
         std::wstring val;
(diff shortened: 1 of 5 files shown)


Samples/CustomTensorization/CustomTensorization/TensorConvertor.cpp
```diff
@@ -33,8 +33,14 @@ namespace TensorizationHelper
 
         val = drive;
         val += dir;
-        int32_t i = val.find(filename);
-        val = val.substr(0, i + wcslen(filename)) + L"\\" + filename + L"\\";
+		if (visiblePath) {
```
youngkim93: tabs?

PR #167, opened 2019-02-19
Add Concurrency scenario

youngkim93:
This change is to add additional options 'ConcurrentLoad' option to load a model multiple times with different threads concurrently. You can also specify two additional parameters 'NumThreads' and 'ThreadInterval' for concurrency options.

To support this scenario we are creating a separate lib and a header file to load model concurrently. 
Also fixing a few nit issues in the source.


Tools/WinMLRunner/WinMLRunner.sln
```diff
@@ -57,6 +59,16 @@ Global
 		{C3BCBEA1-90E6-426F-88AC-64C274BCEF45}.Release|x64.Build.0 = Release|x64
 		{C3BCBEA1-90E6-426F-88AC-64C274BCEF45}.Release|x86.ActiveCfg = Release|Win32
 		{C3BCBEA1-90E6-426F-88AC-64C274BCEF45}.Release|x86.Build.0 = Release|Win32
+		{C174D45D-C189-475B-B1A7-494939EE7491}.Debug|ARM64.ActiveCfg = Debug|Win32
```
ryanlai2: This will build Win32 binary when ARM64 is specified and static linking will be throw errors

Tools/WinMLRunner/WinMLRunner.sln
```diff
@@ -57,6 +59,16 @@ Global
...
+		{C174D45D-C189-475B-B1A7-494939EE7491}.Debug|x64.ActiveCfg = Debug|x64
+		{C174D45D-C189-475B-B1A7-494939EE7491}.Debug|x64.Build.0 = Debug|x64
+		{C174D45D-C189-475B-B1A7-494939EE7491}.Debug|x86.ActiveCfg = Debug|Win32
+		{C174D45D-C189-475B-B1A7-494939EE7491}.Debug|x86.Build.0 = Debug|Win32
+		{C174D45D-C189-475B-B1A7-494939EE7491}.Release|ARM64.ActiveCfg = Release|Win32
```
ryanlai2: Same thing over here :)

Tools/WinMLRunner/src/CommandLineArgs.cpp
```diff
@@ -213,7 +218,21 @@ CommandLineArgs::CommandLineArgs(const std::vector<std::wstring>& args)
...
             return;
         }
-    }
+        else if ((_wcsicmp(args[i].c_str(), L"-ConcurrentLoad") == 0))
+        {
+            ToggleConcurrentLoad(true);
+        }
+        else if ((_wcsicmp(args[i].c_str(), L"-NumThreads") == 0))
+        {
+            unsigned num_threads = std::stoi(args[++i].c_str());
```
ryanlai2: If the user specifies -NumThreads without the required parameter, I think we would need to throw an error saying that the arguments are incorrect.

For example:
./winmlrunner -model ./squeezenet -ConcurrentLoad -NumThreads -CPU  should print an error indicating that there should be an integer argument next to NumThreads flag

Tools/WinMLRunner/src/CommandLineArgs.cpp
```diff
@@ -213,7 +218,21 @@ CommandLineArgs::CommandLineArgs(const std::vector<std::wstring>& args)
...
+            SetNumThreads(num_threads);
+        }
+        else if ((_wcsicmp(args[i].c_str(), L"-ThreadInterval") == 0))
+        {
+            unsigned thread_interval = std::stoi(args[++i].c_str());
```
ryanlai2: Same thing here, I think we should add an extra check to check that the next argument is not another flag but an integer.

Tools/WinMLRunner/WinMLRunnerScenarios.vcxproj
```diff
@@ -0,0 +1,171 @@
...
+    <Link>
+      <SubSystem>Windows</SubSystem>
+      <EnableCOMDATFolding>true</EnableCOMDATFolding>
+      <OptimizeReferences>true</OptimizeReferences>
+      <GenerateDebugInformation>true</GenerateDebugInformation>
+    </Link>
+  </ItemDefinitionGroup>
+  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <ClCompile>
+      <PrecompiledHeader>Use</PrecompiledHeader>
```
ryanlai2: Did you mean to turn PCH off for x64 release?

Tools/WinMLRunner/WinMLRunner.sln
youngkim93: Hmm I wonder why this is happening with visual studio :(

Tools/WinMLRunner/src/CommandLineArgs.cpp
youngkim93: Agreed. We actually have this problem with other parameters too? I think it would be good for us to throw an exception in this function when such cases come up.

ryanlai2: Yes we have this problem with other parameters too. I agree we should throw an exception

Tools/WinMLRunner/WinMLRunner.sln
```diff
@@ -57,6 +59,16 @@ Global
 		{C3BCBEA1-90E6-426F-88AC-64C274BCEF45}.Release|x64.Build.0 = Release|x64
 		{C3BCBEA1-90E6-426F-88AC-64C274BCEF45}.Release|x86.ActiveCfg = Release|Win32
 		{C3BCBEA1-90E6-426F-88AC-64C274BCEF45}.Release|x86.Build.0 = Release|Win32
+		{C174D45D-C189-475B-B1A7-494939EE7491}.Debug|ARM64.ActiveCfg = Debug|ARM64
+		{C174D45D-C189-475B-B1A7-494939EE7491}.Debug|x64.ActiveCfg = Debug|x64
```
ryanlai2: Can we add a ARM64.Build.0 = Debug|ARM64 line here? This way it is build when attempting to build the solution

Tools/WinMLRunner/WinMLRunner.sln
```diff
@@ -57,6 +59,16 @@ Global
...
+		{C174D45D-C189-475B-B1A7-494939EE7491}.Debug|x64.Build.0 = Debug|x64
+		{C174D45D-C189-475B-B1A7-494939EE7491}.Debug|x86.ActiveCfg = Debug|Win32
+		{C174D45D-C189-475B-B1A7-494939EE7491}.Debug|x86.Build.0 = Debug|Win32
+		{C174D45D-C189-475B-B1A7-494939EE7491}.Release|ARM64.ActiveCfg = Release|ARM64
```
ryanlai2: Same thing here. Can we add a ARM64.Build.0 = Release|ARM64 line here?

Tools/WinMLRunner/src/CommandLineArgs.cpp
ryanlai2: Yep yep! Agreed, we should throw exceptions for this function for other cases as well. Thanks Young!

Tools/WinMLRunner/src/CommandLineArgs.cpp
```diff
@@ -34,6 +34,11 @@ void CommandLineArgs::PrintUsage() {
     std::cout << "  -Debug: print trace logs" << std::endl;
     std::cout << "  -Terse: Terse Mode (suppresses repetitive console output)" << std::endl;
     std::cout << "  -AutoScale <interpolationMode>: Enable image autoscaling and set the interpolation mode [Nearest, Linear, Cubic, Fant]" << std::endl;
+    std::cout << std::endl;
+    std::cout << "Concurrency Options:" << std::endl;
```
ryanlai2: Can we specify the usage of concurrency options? Will they ever interact with the normal options? Or is ConcurrentLoad a special case where it shouldn't bind / create session / evaluate? In this case we should specify in the -ConcurrentLoad usage.

Up top on line 12 we already specify: 
WinmlRunner.exe <-model | -folder> <fully qualified path> [options]

youngkim93: I think in the subsequent changes we should also create sesison/bind/evaluate with concurrency options. For now this change is meant to be a boilerplate code for future additions of concurrency options.

ryanlai2: Okay sounds good!

PR #172, opened 2019-02-21
Pass in second value for -SaveTensorData parameter 

youngkim93:
- Pass in second value for -SaveTensorData parameter 
- fix test cases that was not reading csv files correctly


Tools/WinMLRunner/src/CommandLineArgs.cpp
```diff
@@ -191,6 +193,21 @@ CommandLineArgs::CommandLineArgs(const std::vector<std::wstring> &args)
                 PrintUsage();
                 throw hresult_invalid_argument(L"Unknown Mode!");
             }
+            try
```
ryanlai2: On line 9, can we change the usage to:

-SaveTensorData <saveMode> <optional: folder path>: save first iteration or all iteration output tensor results to csv file [First, All]". Optional folder path can be specified to hold tensor data. It will be created if folder doesn't exist.

ryanlai2:
Just to document what still needs to be improved after this PR
- The naming of the tensor files need to include more information about which model and which input and which device it is the result of.
- The format of the tensor files is confusing to read. Maybe for each channel of the output, there could be a column?

ryanlai2:
Also, on the CI build, x86 release needs to build correctly :)

PR #176, opened 2019-02-26
Process FP16 output and output to Console and CSV file

ryanlai2:
This change primarily addresses these main points:
- Before this change, we access the output tensor twice and it was inefficient and had code bloat. It accessed it once to dump the output tensor to CSV and another time to print the max value to console. With this change, there is only one method to access the output tensor once.  PrintOrSaveEvaluationResults is a helper method which controls the cases for the output tensorkind and it calls ProcessTensorResult which is a templated method that handles the access of the output tensor depending on the data type.

- This change fixes the CompareTensor helper method inside the tests so that the comparing or the tensor values is more fair. Epsilon is used here to determine the threshold where the expected tensor value and actual tensor value are too different.

- This change adds squeezenet fp16 model. It was converted with WinMLTools 1.3.0. 

- Output Tensor for FP16 tests and test data were added as well.


Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
```diff
@@ -73,43 +75,44 @@ namespace WinMLRunnerTest
...
+        float epsilon = 0.003f;
+        while (!(expectedFileStream.eof() || actualFileStream.eof()))
+        {
+            std::getline(expectedFileStream, expectedValue, ',');
+            std::getline(expectedFileStream, expectedValue, '\n');
+            std::getline(actualFileStream, actualValue, ',');
+            std::getline(actualFileStream, actualValue, '\n');
+            float actualValueNum = 0;
+            float expectedValueNum = 0;
+            try
```
fdwr: Check `if empty`

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
```diff
@@ -73,43 +75,44 @@ namespace WinMLRunnerTest
...
-                i++;
-                if (i == 1 || i == 1002) continue;
-                if (abs(std::stof(trueValue) - std::stof(value)) > 0.01)
-                {
-                    check = false;
-                    break;
-                }
+                //If the string values aren't floats then ignore
+            }
+            if (std::abs(actualValueNum - expectedValueNum) > epsilon * std::abs(expectedValueNum) + epsilon) //Check if the values are too different.
```
fdwr: relative tolerance is the typical term.

Tools/WinMLRunner/src/OutputHelper.h
```diff
@@ -667,24 +688,108 @@ class OutputHelper
     }
 
     template<typename T>
-    void WriteTensorResultToCSV(T* m_Res, uint32_t iterationNum, const CommandLineArgs &args, uint32_t &capacity, std::string &featureName)
+    void ProcessTensorResult(com_ptr<ITensorNative> itn,
```
fdwr: Try to detemplatize so there isn't code repetition. You can just use a switch statement in the innermost loop.

ryanlai2: Removed out most of the code that's unnecessary for being inside the template. Case switch statement for tensorkind will have to exist somewhere. I chose to have it in the helper method that calls ProcessTensorResult and leave ProcessTensorResult as template for using typename T for looping through the buffer.

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
ryanlai2: Fixed

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
ryanlai2: Fixed

PR #179, opened 2019-02-27
Remove DXCore from WInMLRunner

ryanlai2:
DXCore is no longer contained inside the Preview SDK so this functionality needs to be removed. With this change, WinMLRunner can now be built with (October 2017 Release SDK 17763)


Tools/WinMLRunner/WinMLRunner.vcxproj
```diff
@@ -148,7 +147,6 @@
       <WarningLevel>Level3</WarningLevel>
       <Optimization>Disabled</Optimization>
       <SDLCheck>true</SDLCheck>
-      <ShowIncludes>true</ShowIncludes>
```
youngkim93: What is this for?

ryanlai2: https://docs.microsoft.com/en-us/cpp/build/reference/showincludes-list-include-files?view=vs-2017

PR #181, opened 2019-02-28
Made debug flag documentation more explicit

ryanlai2:
Changed -Debug flag to be -DebugEvaluate so that it's more explicit on what the flag means.

-DebugEvaluateOutput: Print evaluation debug output to debug console if debugger is present.

When this flag is specified and WinMLRunner is under debugger, then EvaluationOptions will be set so that Debug output spew will be printed out into the debug console. The creation of CommandLineArgs will fail and print an error if this flag is specified and WinMLRunner isn't under a debugger.

Also this change updates the README with the latest option of flags.


hahnicity:
Is it possible to update the documentation on how to use this feature? The README still displays usage of the original -Debug flag

PR #185, opened 2019-03-06
Add clang format

youngkim93:
Refactor WinMLRunner using clang format. You can use any of the clang format extension and clang-format file to reformat the code


ryanlai2:
There's a build error in the CI build

PR #186, opened 2019-03-06
Run on all devices even if other device types fail

ryanlai2:
This PR pulls out the code in EvaluateModels that loops through all inputBindingTypes  inputDataTypes and deviceCreationLocations into a separate function.

EvaluateModels loops through all deviceTypes and calls EvaluateModelGivenDeviceType


Tools/WinMLRunner/src/Run.cpp
```diff
@@ -392,65 +448,21 @@ HRESULT EvaluateModels(const std::vector<std::wstring>& modelPaths, const std::v
                        << std::endl;
             continue;
         }
-
         for (const auto& deviceType : deviceTypes)
         {
-            for (const auto& inputBindingType : inputBindingTypes)
+            HRESULT evaluateModelGivenDeviceTypeResult =
+                EvaluateModelGivenDeviceType(model, deviceType, inputBindingTypes, inputDataTypes,
```
youngkim93: Can we rename this to EvaluateModelWithDeviceType?

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -392,65 +448,21 @@ HRESULT EvaluateModels(const std::vector<std::wstring>& modelPaths, const std::v
...
-                                                                 deviceTypeStringified, inputDataTypeStringified,
-                                                                 inputBindingTypeStringified,
-                                                                 deviceCreationLocationStringified);
-                            }
-                        }
-                    }
-                }
+                lastEvaluateModelResult = evaluateModelGivenDeviceTypeResult;
+                std::cout << "Run failed for DeviceType: " << TypeHelper::Stringify(deviceType) << std::endl;
+                ;
```
youngkim93: nit: extra ; :)

Tools/WinMLRunner/src/Run.cpp
ryanlai2: Sure sure! I can rename this.

Tools/WinMLRunner/src/Run.cpp
ryanlai2: Good catch!

youngkim93:
It would be great if we can add test cases for this, but I understand that this might be hard to repro (since you should purposely fail in one of the devices, which is not what where should be at all)

ryanlai2:
> It would be great if we can add test cases for this, but I understand that this might be hard to repro (since you should purposely fail in one of the devices, which is not what where should be at all)

Right, it would be great to add test cases for this. However, it would be hard to find a model that purposely fails on one of the devices but passes on others. 

In the future, I believe using mocking framework would be best to test this.

PR #189, opened 2019-03-06
Printing top K values instead of just single max value

youngkim93:
Currently WinMLRunner prints top value in the result float tensor. This change is to output top K values instead: 

Here is the sample output:

Creating Session with GPU: AMD FirePro W4100 Graphics Adapter
Binding (device = GPU, iteration = 1, inputBinding = CPU, inputDataType = RGB_Image, deviceCreationLocation = WinML)...[SUCCESS]
Outputting top 5 values
Feature Name: resnetv23_dense0_fwd
    index: 409, value: 6.74914
    index: 769, value: 5.66233
    index: 664, value: 5.62002
    index: 585, value: 5.48044
    index: 872, value: 5.29839
Evaluating (device = GPU, iteration = 1, inputBinding = CPU, inputDataType = RGB_Image, deviceCreationLocation = WinML)...[SUCCESS]

### DIFF
--- a/Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
+++ b/Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
@@ -749,6 +749,12 @@ namespace WinMLRunnerTest
             Assert::AreEqual(S_OK, RunProc((wchar_t*)command.c_str()));
         }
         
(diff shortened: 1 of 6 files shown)


ryanlai2:
Looks good to me! Can you update the usage in README.md to show this flag as well?

PR #194, opened 2019-03-12
azure pipelines fixes

orilevari:
1) remove build variables from yaml which now are causing early error in the build pipeline. Seems as if it is due to a change in azure pipelines behavior: https://mscodehub.visualstudio.com/WindowsAI/_build/results?buildId=85261&view=results
2) Also add pull request validation and change filtering of paths to just exclude the Tools directory since it is only the Tools directory which is properly testing by other pipelines

### DIFF
--- a/azure-pipelines-samples.yml
+++ b/azure-pipelines-samples.yml
@@ -6,10 +6,7 @@
 variables:
 
   SamplesBin: SamplesBin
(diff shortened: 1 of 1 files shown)


orilevari:
This pull request still failed to trigger a PR build, but I think this is just because its not actually checked in to master yet.

PR #195, opened 2019-03-12
User/orilevari/normalizeappxpublishers

orilevari:
Change all sample store apps to have the same publisher (CN=winml) so that during a test pipelines we can use 1 self-signed certificate that matches the publisher of all of our apps.

### DIFF
--- a/Samples/FNSCandyStyleTransfer/UWP/cs/Package.appxmanifest
+++ b/Samples/FNSCandyStyleTransfer/UWP/cs/Package.appxmanifest
@@ -4,7 +4,7 @@
   <mp:PhoneIdentity PhoneProductId="3bd922b5-d779-4208-8c35-aec4dea61330" PhonePublisherId="00000000-0000-0000-0000-000000000000" />
   <Properties>
     <DisplayName>Shortcut</DisplayName>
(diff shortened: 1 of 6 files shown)


Samples/FNSCandyStyleTransfer/UWP/cs/Package.appxmanifest
```diff
@@ -1,10 +1,10 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Package xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10" xmlns:mp="http://schemas.microsoft.com/appx/2014/phone/manifest" xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10" IgnorableNamespaces="uap mp">
-  <Identity Name="257918C9-ABE3-483E-A202-C5C69AEBD825" Publisher="CN=lobourre" Version="1.1.36.0" />
+  <Identity Name="257918C9-ABE3-483E-A202-C5C69AEBD825" Publisher="CN=winml" Version="1.1.36.0" />
```
youngkim93: Does version number also have to be consistent across apps?

PR #198, opened 2019-03-14
User/paulm/fps

walrusmcd:
added counters to keep track of capture FPS and render FPS

### DIFF
--- a/Samples/FNSCandyStyleTransfer/UWP/cs/MainPage.xaml.cs
+++ b/Samples/FNSCandyStyleTransfer/UWP/cs/MainPage.xaml.cs
@@ -80,6 +80,10 @@ namespace SnapCandy
 
         // Debug
         private Stopwatch _perfStopwatch = new Stopwatch(); // performance Stopwatch used throughout
+        private DispatcherTimer _FramesPerSecondTimer = new DispatcherTimer();
+        private long _CaptureFPS = 0;
+        public static long _RenderFPS = 0;
+        private int _LastFPSTick = 0;
 
         public MainPage()
         {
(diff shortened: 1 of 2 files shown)


Samples/FNSCandyStyleTransfer/UWP/cs/MainPage.xaml.cs
```diff
@@ -110,6 +114,39 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
...
+            _FramesPerSecondTimer.Tick += _FramesPerSecond_Tick;
+            _FramesPerSecondTimer.Interval = new TimeSpan(0, 0, 1);
+            _FramesPerSecondTimer.Start();
+        }
+
+        private void _FramesPerSecond_Tick(object sender, object e)
+        {
+            // how many seconds has it been?
+            int fpsTick = System.Environment.TickCount;
+            float numberOfSeconds = ((float)(fpsTick - _LastFPSTick)) / (float)1000;
```
ryanlai2: Seems to me like the first numberOfSeconds will be incorrect since _LastFPSTick is initialized to 0. 

However, since _LastFPSTick is updated on line 144, then it will be correct for future calls to _FramesPerSecond_Tick.

Samples/FNSCandyStyleTransfer/UWP/cs/MainPage.xaml.cs
```diff
@@ -110,6 +114,39 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
...
+
+            // Create a 1 second timer
+            _FramesPerSecondTimer.Tick += _FramesPerSecond_Tick;
+            _FramesPerSecondTimer.Interval = new TimeSpan(0, 0, 1);
+            _FramesPerSecondTimer.Start();
+        }
+
+        private void _FramesPerSecond_Tick(object sender, object e)
+        {
+            // how many seconds has it been?
```
ryanlai2: We calculate seconds in _FramesPerSecond_tick because even though this method is queued every second by DispatcherTimer, it doesn't mean that it will be called in exactly every second right?

Samples/FNSCandyStyleTransfer/UWP/cs/MainPage.xaml.cs
walrusmcd: fixed !

Samples/FNSCandyStyleTransfer/UWP/cs/MainPage.xaml.cs
walrusmcd: i added more comments.   you are correct !

PR #203, opened 2019-03-25
Various bug fixes and build features to public samples

orilevari:
This pull request includes a handle of bug fixes and build configuration changes that together comprise an effort to automate build and test pipelines for our public samples. Here is a list of the changes included more or less in the order the appear in the files changed tab:

1) Change the manner in which user input is consumed in the adapter selection sample in order to avoid an infinite loop bug that manifested which running on the release pipeline.

2) Modify the manner in which the CustomTensorization sample would try to output files to disk. The intended behavior of the sample is that if it is ran within the directory structure of this repository (the standard use case) it will output files into the directory containing the sample source code. Otherwise it will be output next to the executable. However, the path resolution logic would fail in certain circumstances that were not considered until a separate change to the organization of our build artifacts (see below).

3) Include temporary pfx files into the repository in order to build our UWP apps remotely with appx signing enabled. Appx signing needs to be enabled for us to be able to deploy our UWP apps on a remote server and leverage the existing Add-AppDevPackage.ps1 script.

4) Rename the publisher of our UWP apps to winml for simplicity's sake.

5) Avoid autoincrementing our UWP msix packages in order to keep our build artifacts consistent over time.

6) Avoid bundling into msix bundles so that we can be sure we are installing and testing a specific Build Platform flavor in our test pipeline.

7) Correct the Mnist_cppcx mapping of BuildPlatform for Debug|x64

8) Refactor the UI SamplesTest to leverage the same Setup code. Modifies this setup code to fallback to a secondary launch option in the case where the application takes too long to launch. Also update the AppIds to reflect the modified publishers explained above.

9) Modify our build pipeline to enable appx signing

10) Modify our build pipeline to organize our build artifacts by sample and exclude unnecessary build artifacts.

11) Modify the StyleTransferTest to compare images using standard C# apis rather than a raw byte comparison, since images of the same dimensions can differ in byte encoding. This was manifesting when building and testing the samples on machines with differing configurations.


orilevari:
Note the name of the branch is not indicative of the breadth of changes included in this pull request

Testing/SamplesTest/SamplesTest/StyleTransferTest.cs
```diff
@@ -162,5 +178,7 @@ public byte[] ImageToByteArray(string imagePath)
             img.Save(ms, img.RawFormat);
             return ms.ToArray();
         }
+
```
youngkim93: nit: extra spaces

youngkim93:
What are the differences in the new baseline .png files?

PR #204, opened 2019-03-26
Modify debug_all_outputs.py

orilevari:
This pull request changes the debug_all_outputs.py helper script to not insert a debug custom operator after any final graph outputs, because this was violating the onnx standard and causing a bug due to undefined behavior

### DIFF
--- a/Samples/CustomOperatorCPU/customize_model/scripts/debug_all_outputs.py
+++ b/Samples/CustomOperatorCPU/customize_model/scripts/debug_all_outputs.py
@@ -18,7 +18,8 @@ def create_modified_model(args):
         for output in node.output:
             intermediate_outputs.add(output)
 
-    # remove final graph outputs
+    # remove final graph outputs (see the definition of graph outputs in the onnx standard: https://github.com/onnx/onnx/blob/master/docs/IR.md#user-content-graphs)
+    # it would be invalid to add debug operators after the final graph outputs since execution must be finished once all outputs are written to
     for output in model.graph.output:
         intermediate_outputs.remove(output.name)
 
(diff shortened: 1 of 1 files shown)


Samples/CustomOperatorCPU/customize_model/scripts/debug_all_outputs.py
```diff
@@ -18,6 +18,10 @@ def create_modified_model(args):
         for output in node.output:
             intermediate_outputs.add(output)
 
+    # remove final graph outputs
```
walrusmcd: so we don't want the graph outputs to be debugged?

Samples/CustomOperatorCPU/customize_model/scripts/debug_all_outputs.py
```diff
@@ -18,6 +18,10 @@ def create_modified_model(args):
         for output in node.output:
             intermediate_outputs.add(output)
 
+    # remove final graph outputs
+    for output in model.graph.output:
```
walrusmcd: so we don't want the graph outputs to be debugged?

Samples/CustomOperatorCPU/customize_model/scripts/debug_all_outputs.py
youngkim93: Which spec in onnx are we violating? Can we add a link as a comment here?

orilevari: The final graph outputs are already available as the results of the evaluation. Turns out that the onnx specification defines the execution of a graph to be finished once all the graph outputs are completed. Therefore inserting debug operators after the graph outputs is introducing undefined behavior. And doesn't add anything when it comes to debugging. I would say these are invalid models.

However, onnxruntime should fail gracefully when running this kind of model, when in reality it works some of the time and then happened to crash when running on a model with FP16.

Either way, my helper script shouldn't output this extra debug operator either way, and that is what this pull request addresses.

And yes I will add a link in a comment

walrusmcd: ahh... this answers my question below.  thx !

PR #209, opened 2019-04-09
Support Models with Image Denotation Input

ryanlai2:
This change is to add support with models for image denotation.

There was refactor work in how the code handles features descriptors. 

There was also work to remove ModelBinding.h entirely. With this change, CSV tensors are written into tensor binding with a zero copy strategy.

A function wide "try / catch" was added to the main run method to catch the most common error messages.


Tools/WinMLRunner/src/BindingUtilities.h
```diff
@@ -127,26 +126,46 @@ template <> struct TensorKindToValue<TensorKind::String>
...
+    }
+}
+
 namespace BindingUtilities
 {
     static unsigned int seed = 0;
     static std::independent_bits_engine<std::default_random_engine, CHAR_BIT, unsigned int> randomBitsEngine;
 
-    SoftwareBitmap GenerateGarbageImage(const TensorFeatureDescriptor& imageDescriptor, InputDataType inputDataType)
+    SoftwareBitmap GenerateGarbageImage(const ILearningModelFeatureDescriptor& modelFeatureDescriptor,
```
youngkim93: It seems the function is assuming the input descriptor is imagefeature. Should we narrow the scope of this input parameter to just accept ImageFeatureDescriptor?

Tools/WinMLRunner/src/BindingUtilities.h
```diff
@@ -168,17 +187,16 @@ namespace BindingUtilities
                                                     static_cast<int32_t>(width), static_cast<int32_t>(height));
     }
 
-    SoftwareBitmap LoadImageFile(const TensorFeatureDescriptor& imageDescriptor, InputDataType inputDataType,
-                                 const hstring& filePath, const CommandLineArgs& args, uint32_t iterationNum)
+    SoftwareBitmap LoadImageFile(const ILearningModelFeatureDescriptor& modelFeatureDescriptor,
```
youngkim93: Same here. It looks like both of these functions are called from CreateBindableImage, and I think it's better to do type check there before we call these functions

Tools/WinMLRunner/src/BindingUtilities.h
```diff
@@ -269,18 +287,26 @@ namespace BindingUtilities
...
+        if (bindingMemorySize / sizeof(T) != elementStrings.size())
         {
             throw hresult_invalid_argument(L"CSV Input is size/shape is different from what model expects");
         }
-        T* data = binding.GetData();
+        T* data = bindingMemory;
         for (const auto& elementString : elementStrings)
         {
-            T value;
+            float value;
```
youngkim93: any reason we are not using a template type here?

youngkim93: Ah I see. Regardless of half or float type we are taking it as a float.

Tools/WinMLRunner/src/BindingUtilities.h
youngkim93: Never mind this comment as this can be a TensorFeatureDescriptor

Tools/WinMLRunner/src/BindingUtilities.h
```diff
@@ -300,142 +326,161 @@ namespace BindingUtilities
...
-            BYTE* actualData;
-            uint32_t actualSizeInBytes;
-            spTensorValueNative->GetBuffer(
-                &actualData, &actualSizeInBytes); // Need to GetBuffer to have CPU memory backing tensorValue
+        BYTE* actualData;
+        uint32_t actualSizeInBytes;
+        spTensorValueNative->GetBuffer(&actualData,
+                                       &actualSizeInBytes); // Need to GetBuffer to have CPU memory backing tensorValue
+
+        if (!args.CsvPath().empty())
```
youngkim93: args.IsCsvInput()? Also maybe args.IsImageInput()?

Tools/WinMLRunner/src/BindingUtilities.h
```diff
@@ -300,142 +326,161 @@ namespace BindingUtilities
...
+        {
+            WriteDataToBinding<DataType>(tensorStringInput, reinterpret_cast<DataType*>(actualData), actualSizeInBytes);
+        }
+        else if (args.IsGarbageInput())
+        {
             return tensorValue;
         }
```
youngkim93: remove this else if and add else if (args.IsImage()) { ... } return tensorValue;

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -79,7 +78,7 @@ std::vector<ILearningModelFeatureValue> GenerateInputFeatures(const LearningMode
 
         if (inputDataType == InputDataType::Tensor || i > 0)
```
youngkim93: remove i > 0

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -436,6 +426,64 @@ HRESULT EvaluateModelWithDeviceType(const LearningModel& model, const DeviceType
     return S_OK;
 }
 
+HRESULT CheckIfModelAndConfigurationsAreSupported(LearningModel& model, const std::wstring& modelPath,
+                                                  const DeviceType& deviceType,
```
youngkim93: Don't really need reference since it's  enum

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -436,6 +426,64 @@ HRESULT EvaluateModelWithDeviceType(const LearningModel& model, const DeviceType
...
+                                                  const std::vector<InputDataType>& inputDataTypes,
+                                                  const std::vector<DeviceCreationLocation> deviceCreationLocations)
```
youngkim93: &

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -436,6 +426,64 @@ HRESULT EvaluateModelWithDeviceType(const LearningModel& model, const DeviceType
...
+            if (tensorFeatureDescriptor.Shape().Size() > 4 && deviceType != DeviceType::CPU)
+            {
+                std::cout << "Input feature " << to_string(inputFeature.Name())
+                          << " shape is too large. GPU path only accepts tensor dimensions <= 4 : "
+                          << tensorFeatureDescriptor.Shape().Size() << std::endl;
+                return E_INVALIDARG;
+            }
+
+            // If image as input binding, then the model's tensor inputs should have channel 3 or 1
+            if (isInputBindingImage && tensorFeatureDescriptor.Shape().Size() == 4 &&
```
youngkim93: if (isInputBindingImage) 
{
 if (tensorFeatureDescriptor.Shape().size() != 4 || (tensorFeatureDescriptor.Shape().GetAt(1) != 1 && tensorFeatureDescriptor.Shape().GetAt(1) != 3))

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -436,6 +426,64 @@ HRESULT EvaluateModelWithDeviceType(const LearningModel& model, const DeviceType
...
+                std::cout << "Attempting to bind image but input feature " << to_string(inputFeature.Name())
+                          << " shape is invalid. Shape should be 4 dimensions (NCHW) with C = 3." << std::endl;
+                return E_INVALIDARG;
+            }
+        }
+    }
+
+    // Creating D3D12 device on client doesn't make sense for CPU deviceType
+    if (deviceType == DeviceType::CPU && std::find(deviceCreationLocations.begin(), deviceCreationLocations.end(),
+                                                   DeviceCreationLocation::ClientCode) != deviceCreationLocations.end())
```
youngkim93: Let's rename ClientCode to UserD3DDevice?

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -436,6 +426,64 @@ HRESULT EvaluateModelWithDeviceType(const LearningModel& model, const DeviceType
     return S_OK;
 }
 
+HRESULT CheckIfModelAndConfigurationsAreSupported(LearningModel& model, const std::wstring& modelPath,
+                                                  const DeviceType& deviceType,
+                                                  const std::vector<InputDataType>& inputDataTypes,
+                                                  const std::vector<DeviceCreationLocation> deviceCreationLocations)
+{
+    // Does user want image as input binding
+    bool isInputBindingImage = false;
```
youngkim93: rename to hasInputBindingImage and try std::any_of?

PR #211, opened 2019-04-10
Refactor code for winmlrunner

youngkim93:
This change is to refactor codes so that we can easily identify which code section is doing "load | Session | Bind | Eval " part. This will be later used to specify options to do certain operations only for performance measurement.


Tools/WinMLRunner/src/Run.cpp
```diff
@@ -136,69 +80,46 @@ HRESULT BindInputFeatures(const LearningModel& model, const LearningModelBinding
...
     return S_OK;
 }
 
-// Binds and evaluates the user-specified model and outputs success/failure for each step. If the
-// perf flag is used, it will output the CPU, GPU, and wall-clock time for each step to the
-// command-line and to a CSV file.
-HRESULT EvaluateModel(const LearningModel& model, const CommandLineArgs& args, OutputHelper& output,
-                      DeviceType deviceType, InputBindingType inputBindingType, InputDataType inputDataType,
-                      DeviceCreationLocation deviceCreationLocation, Profiler<WINML_MODEL_TEST_PERF>& profiler)
+HRESULT CreateSession(LearningModelSession& session, IDirect3DDevice &winrtDevice, LearningModel& model, CommandLineArgs& args, OutputHelper& output,
```
ryanlai2: nit: IDirect3DDevice& winrtDevice

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -285,147 +206,73 @@ HRESULT EvaluateModel(const LearningModel& model, const CommandLineArgs& args, O
         session.EvaluationProperties().Insert(L"EnableDebugOutput", nullptr);
     }
 
-    LearningModelBinding context(session);
+    return S_OK;
+}
 
+HRESULT BindInputs(LearningModelBinding &context, const LearningModel& model, const LearningModelSession& session, OutputHelper& output,
+                   DeviceType deviceType, const CommandLineArgs& args, InputBindingType inputBindingType, InputDataType inputDataType,
+                   const IDirect3DDevice winrtDevice, DeviceCreationLocation deviceCreationLocation, uint32_t iteration,
```
ryanlai2: const IDirect3DDevice winrtDevice should be pass by reference

PR #214, opened 2019-04-16
Fix x86 missing include header

ryanlai2:
Forgot to include header for x86 configuration

### DIFF
--- a/Tools/WinMLRunner/src/Run.cpp
+++ b/Tools/WinMLRunner/src/Run.cpp
@@ -532,6 +532,12 @@ int run(CommandLineArgs& args, Profiler<WINML_MODEL_TEST_PERF>& profiler) try
                     {
                        for (auto inputBindingType : inputBindingTypes)
                        {
+                           // Clear up bind, eval performance metrics after iteration
+                           if (args.IsPerformanceCapture() || args.IsPerIterationCapture())
+                           {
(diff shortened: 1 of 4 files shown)


Tools/WinMLRunner/src/Run.cpp
```diff
@@ -532,6 +532,12 @@ int run(CommandLineArgs& args, Profiler<WINML_MODEL_TEST_PERF>& profiler) try
                     {
                        for (auto inputBindingType : inputBindingTypes)
                        {
+                           // Clear up bind, eval performance metrics after iteration
+                           if (args.IsPerformanceCapture() || args.IsPerIterationCapture())
+                           {
+                               // Resets all values from profiler for bind and evaluate.
```
ryanlai2: I piggy backed this PR for this change: which resets Bind and Evaluate profiling values before executing all the iterations. The reason for putting this up here is so that the profiler doesn't get reset after collecting perf numbers for bind + eval before returning. 

For code that consumes WinMLRunner as a static library, with the reset at the bottom, the profiler gets reset right before "run" returns.

youngkim93: I see so this change is to capture performance metrics from the profiler after run call right?

ryanlai2: Yep yep!

PR #216, opened 2019-04-18
Add -BaseOutputDir and more robust path support in general.  Fix PerI…

pmbrown1055:
…terationPerf issue with Unit Test.

### DIFF
--- a/Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
+++ b/Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
@@ -38,7 +38,7 @@ namespace WinMLRunnerTest
     static const std::wstring EXE_PATH = CURRENT_PATH + L"WinMLRunner.exe";
     static const std::wstring INPUT_FOLDER_PATH = CURRENT_PATH + L"test_folder_input";
     static const std::wstring OUTPUT_PATH = CURRENT_PATH + L"test_output.csv";
-    static const std::wstring TENSOR_DATA_PATH = CURRENT_PATH + L"\\TestResult";
+    static const std::wstring TENSOR_DATA_PATH = CURRENT_PATH + L"TestResult";
 
     static std::wstring BuildCommand(std::initializer_list<std::wstring>&& arguments)
     {
@@ -52,11 +52,16 @@ namespace WinMLRunnerTest
         return commandLine;
     }
 
-    static size_t GetOutputCSVLineCount()
+    static size_t GetOutputCSVLineCount(const std::wstring& path)
     {
         std::ifstream fin;
-        fin.open(OUTPUT_PATH);
+        fin.open(path);
         return static_cast<size_t>(std::count(std::istreambuf_iterator<char>(fin), std::istreambuf_iterator<char>(), '\n'));
+    }
+
+    static size_t GetOutputCSVLineCount()
+    {
+        return GetOutputCSVLineCount(OUTPUT_PATH);
     }
 
     static void RemoveModelsFromFolder(std::initializer_list<std::string>&& modelList)
@@ -262,7 +267,6 @@ public: TEST_CLASS_INITIALIZE(SetupClass) {
             // We need to expect one more line because of the header
             Assert::AreEqual(static_cast<size_t>(3), GetOutputCSVLineCount());
         }
-
         TEST_METHOD(GarbageInputOnlyCpu)
         {
             const std::wstring modelPath = CURRENT_PATH + L"SqueezeNet.onnx";
@@ -273,7 +277,6 @@ public: TEST_CLASS_INITIALIZE(SetupClass) {
             // We need to expect one more line because of the header
             Assert::AreEqual(static_cast<size_t>(2), GetOutputCSVLineCount());
         }
-
         TEST_METHOD(GarbageInputOnlyGpu)
         {
             const std::wstring modelPath = CURRENT_PATH + L"SqueezeNet.onnx";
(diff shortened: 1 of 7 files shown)


Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
```diff
@@ -52,13 +52,18 @@ namespace WinMLRunnerTest
         return commandLine;
     }
 
-    static size_t GetOutputCSVLineCount()
```
ryanlai2: const and pass by reference for std::wstring path

Tools/WinMLRunner/src/CommandLineArgs.cpp
```diff
@@ -318,6 +318,64 @@ CommandLineArgs::CommandLineArgs(const std::vector<std::wstring>& args)
             throw hresult_invalid_argument(msg.c_str());
         }
     }
+
+    // Set up output directories
+    {
```
ryanlai2: Extra brackets?

Tools/WinMLRunner/src/CommandLineArgs.cpp
```diff
@@ -318,6 +318,64 @@ CommandLineArgs::CommandLineArgs(const std::vector<std::wstring>& args)
...
+        std::filesystem::path PerfOutputPath(sPerfOutputPath);
```
ryanlai2: Do sPerfOutputPath, sBaseOutputPath, and sPerIterationDataPath need to be initialized? Does the std::filesystem::path constructor handle unintialized wstrings?

Tools/WinMLRunner/src/CommandLineArgs.cpp
pmbrown1055: Trying to limit the scope of the filesystem path variables.  Open to removing them.

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -605,6 +598,18 @@ int run(CommandLineArgs& args, Profiler<WINML_MODEL_TEST_PERF>& profiler) try
...
                                 }
                             }
+
+                            if (SUCCEEDED(lastHr) && args.IsPerIterationCapture())
+                            {
+                                output.WritePerIterationPerformance(args, model.Name().c_str());
+                            }
+
+                            // Clear up bind, eval performance metrics after iteration
+                            if (args.IsPerformanceCapture() || args.IsPerIterationCapture())
```
ryanlai2: There is already a reset to the profiler @line 529. If a reset was added here, the profiler object that's passed by reference would be reset before the "run" method returns. It would break users who use WinMLRunner as a static library and would like the profiler object populated when passed into run.

Tools/WinMLRunner/src/CommandLineArgs.cpp
pmbrown1055: Default string constructor sets the string to null string (L""), which I verified in the debugger.  Since they're empty, std::filesystem::path handles the case without a problem.

I guess one thing to note is that the paths are initialized to default values, even if the outputs are not turned on... I err'd on the side of trying to keep the code simpler, and to have the paths always initialized to something, rather than only init them if -perfoutput or -periterationperf or -savetensor is specified.  And it's one more place where future additions would need to be added for more output options...

ryanlai2: Thanks for verifying! Sounds good to me

Tools/WinMLRunner/src/CommandLineArgs.cpp
ryanlai2: I would prefer to have this code separated in a member function to make the code more readable. What do you think?

pmbrown1055: That's fine, I would just be passing in the 3 strings by reference, right?

Tools/WinMLRunner/src/Run.cpp
pmbrown1055: Sorry, this was a merge error.  Removing it.

Tools/WinMLRunner/src/CommandLineArgs.cpp
ryanlai2: Yep yep!

Tools/WinMLRunner/src/CommandLineArgs.cpp
```diff
@@ -319,64 +321,68 @@ CommandLineArgs::CommandLineArgs(const std::vector<std::wstring>& args)
...
-        std::filesystem::path PerfOutputPath(sPerfOutputPath);
-        std::filesystem::path BaseOutputPath(sBaseOutputPath);
-        std::filesystem::path PerIterationDataPath(sPerIterationDataPath);
+    SetupOutputDirectories(sBaseOutputPath, sPerfOutputPath, sPerIterationDataPath);
 
-        if (PerfOutputPath.is_absolute())
+    CheckForInvalidArguments();
+}
+
+void CommandLineArgs::SetupOutputDirectories(const std::wstring &sBaseOutputPath,
```
ryanlai2: nit: const std::wstring& sBaseOutputPath,

Tools/WinMLRunner/src/CommandLineArgs.h
```diff
@@ -98,19 +98,18 @@ class CommandLineArgs
...
+        m_perIterationDataPath = perIterationDataPath;
+    }
     void SetInputDataPath(const std::wstring& inputDataPath) { m_inputData = inputDataPath; }
     void SetNumThreads(unsigned numThreads) { m_numThreads = numThreads; }
     void SetThreadInterval(unsigned threadInterval) { m_threadInterval = threadInterval; }
     void SetTopK(unsigned k) { m_topK = k; }
-    void SetPerformanceCSVPath(const std::wstring& performanceCSVPath)
-    {
-        m_perfOutputPath = performanceCSVPath;
-        m_perfOutput = true;
```
ryanlai2: Since ```m_perfOutput = true``` is getting removed, can you add a small  "TogglePerfOutput(bool perfOutput)" method?

PR #217, opened 2019-04-18
Add back DxCore support to WinMLRunner

ryanlai2:
This change adds back Dxcore to WinMLRunner.

- -GPUAdapterIndex <index number> : run model on GPU specified by its DXCore index. NOTE: Please only use this flag on DXCore supported machines.
- -GPUAdapterName <adapter name substring>: run model on GPU specified by its name. NOTE: Please only use this flag on DXCore supported machines.

Examples:
- ./WinMLRunner -model ./SqueezeNet.onnx -GpuAdapterName vpu
- ./WinMLRunner -model ./SqueezeNet.onnx -GpuAdapterIndex 2

### DIFF
--- a/Tools/WinMLRunner/README.md
+++ b/Tools/WinMLRunner/README.md
@@ -35,7 +35,6 @@ Required command-Line arguments:
 -GPU : run model on default GPU
 -GPUHighPerformance : run model on GPU with highest performance
 -GPUMinPower : run model on GPU with the least power
--GPUAdapterIndex <index number> : run model on GPU specified by its DXCore index. NOTE: Please only use this flag on DXCore supported machines.
 -GPUAdapterName <adapter name substring>: run model on GPU specified by its name. NOTE: Please only use this flag on DXCore supported machines.
 -CreateDeviceOnClient : create the D3D device on the client and pass it to WinML to create session
 -CreateDeviceInWinML : create the device inside WinML
--- a/Tools/WinMLRunner/README.md
+++ b/Tools/WinMLRunner/README.md
@@ -35,6 +35,8 @@ Required command-Line arguments:
 -GPU : run model on default GPU
 -GPUHighPerformance : run model on GPU with highest performance
 -GPUMinPower : run model on GPU with the least power
+-GPUAdapterIndex <index number> : run model on GPU specified by its DXCore index. NOTE: Please only use this flag on DXCore supported machines.
+-GPUAdapterName <adapter name substring>: run model on GPU specified by its name. NOTE: Please only use this flag on DXCore supported machines.
 -CreateDeviceOnClient : create the D3D device on the client and pass it to WinML to create session
 -CreateDeviceInWinML : create the device inside WinML
 -CPUBoundInput : bind the input to the CPU
--- a/Tools/WinMLRunner/WinMLRunner.vcxproj
+++ b/Tools/WinMLRunner/WinMLRunner.vcxproj
@@ -160,6 +160,7 @@
       <SubSystem>Console</SubSystem>
       <GenerateDebugInformation>true</GenerateDebugInformation>
       <AdditionalDependencies>WindowsApp.lib; mincore.lib; DXGI.lib</AdditionalDependencies>
+      <DelayLoadDLLs>"ext-ms-win-dxcore-l1-1-0.dll"</DelayLoadDLLs>
     </Link>
     <PreBuildEvent>
       <Command>$(ProjectDir)src\GenerateVersionStrings.cmd $(ProjectDir) $(IntDir)</Command>
@@ -185,6 +186,7 @@
       <SubSystem>Console</SubSystem>
       <GenerateDebugInformation>true</GenerateDebugInformation>
       <AdditionalDependencies>WindowsApp.lib; mincore.lib; DXGI.lib</AdditionalDependencies>
+      <DelayLoadDLLs>"ext-ms-win-dxcore-l1-1-0"</DelayLoadDLLs>
     </Link>
     <ResourceCompile>
       <AdditionalIncludeDirectories>$(IntDir)</AdditionalIncludeDirectories>
@@ -211,6 +213,7 @@
       <SubSystem>Console</SubSystem>
       <GenerateDebugInformation>true</GenerateDebugInformation>
       <AdditionalDependencies>WindowsApp.lib; mincore.lib; DXGI.lib</AdditionalDependencies>
+      <DelayLoadDLLs>"ext-ms-win-dxcore-l1-1-0"</DelayLoadDLLs>
     </Link>
     <ResourceCompile>
       <AdditionalIncludeDirectories>$(IntDir)</AdditionalIncludeDirectories>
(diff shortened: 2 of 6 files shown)


Tools/WinMLRunner/src/Run.cpp
```diff
@@ -209,12 +331,14 @@ HRESULT CreateSession(LearningModelSession& session, IDirect3DDevice& winrtDevic
     return S_OK;
 }
 
-HRESULT BindInputs(LearningModelBinding &context, const LearningModel& model, const LearningModelSession& session, OutputHelper& output,
-                   DeviceType deviceType, const CommandLineArgs& args, InputBindingType inputBindingType, InputDataType inputDataType,
-                   const IDirect3DDevice& winrtDevice, DeviceCreationLocation deviceCreationLocation, uint32_t iteration,
+HRESULT BindInputs(LearningModelBinding& context, const LearningModel& model, const LearningModelSession& session,
```
ryanlai2: Every change below this line is made by the clang formatter

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -178,6 +180,126 @@ HRESULT CreateSession(LearningModelSession& session, IDirect3DDevice& winrtDevic
...
+        {
+             com_ptr<IDXCoreAdapterFactory> spFactory;
+             THROW_IF_FAILED(DXCoreCreateAdapterFactory(IID_PPV_ARGS(spFactory.put())));
+
+             com_ptr<IDXCoreAdapterList> spAdapterList;
+             const GUID dxGUIDs[] = { DXCORE_ADAPTER_ATTRIBUTE_D3D12_CORE_COMPUTE };
+
+             THROW_IF_FAILED(spFactory->GetAdapterList(dxGUIDs, ARRAYSIZE(dxGUIDs), spAdapterList.put()));
+
+             CHAR driverDescription[128];
```
jeffwick-msft: It would be better to dynamical determine the size needed using QueryPropertySize()

Tools/WinMLRunner/WinMLRunner.vcxproj
```diff
@@ -211,6 +213,7 @@
       <SubSystem>Console</SubSystem>
       <GenerateDebugInformation>true</GenerateDebugInformation>
       <AdditionalDependencies>WindowsApp.lib; mincore.lib; DXGI.lib</AdditionalDependencies>
+      <DelayLoadDLLs>"ext-ms-win-dxcore-l1-1-0.dll"</DelayLoadDLLs>
```
fdwr: Hmm, have you asked Jeff W if this is the correct DLL? It's kinda weird for an app to directly link to specifically versioned ext-ms DLL's. (usually only OS components do that)

fdwr: @jeffwick-msft : Is this proper?

Tools/WinMLRunner/README.md
```diff
@@ -35,6 +35,8 @@ Required command-Line arguments:
 -GPU : run model on default GPU
 -GPUHighPerformance : run model on GPU with highest performance
 -GPUMinPower : run model on GPU with the least power
+-GPUAdapterIndex <index number> : run model on GPU specified by its DXCore index. NOTE: Please only use this flag on DXCore supported machines.
```
jeffwick-msft: The adapter indexes are only guaranteed to be the same for the lifetime of the adapter list, and therefore we don't reliably know them ahead of time. I'd lean towards removing this option.

indygit: But it is very convenient, also WinMLRunner prints out the available adapters.

Tools/WinMLRunner/src/CommandLineArgs.cpp
```diff
@@ -19,6 +19,12 @@ void CommandLineArgs::PrintUsage()
     std::cout << "  -GPU : run model on default GPU" << std::endl;
     std::cout << "  -GPUHighPerformance : run model on GPU with highest performance" << std::endl;
     std::cout << "  -GPUMinPower : run model on GPU with the least power" << std::endl;
+#ifdef MCDM_BUILD
```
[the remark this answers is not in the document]
indygit: To ensure compatibility, we actually avoid using DXCore if an adapter is also available from DXGI. :-) Over time, we should move over to DXCore. But I think the macro name change is good.

Tools/WinMLRunner/README.md
jeffwick-msft: True - but there's no guarantee those indexes won't change from execution to execution.

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -178,6 +180,126 @@ HRESULT CreateSession(LearningModelSession& session, IDirect3DDevice& winrtDevic
...
+             std::map<int, com_ptr<IDXCoreAdapter>> validAdapters;
+             printf("Printing available adapters..\n");
+             for (UINT i = 0; i < spAdapterList->GetAdapterCount(); i++)
+             {
+                 com_ptr<IDXCoreAdapter> spAdapter;
+                 THROW_IF_FAILED(spAdapterList->GetItem(i, spAdapter.put()));
+                 // If the adapter is a software adapter then don't consider it
+                 bool isHardware;
+                 THROW_IF_FAILED(spAdapter->QueryProperty(DXCoreProperty::IsHardware, sizeof(isHardware), &isHardware));
+                 if (isHardware)
```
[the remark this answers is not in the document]
ryanlai2: With the current code, there isn't any option to select WARP either with index or GpuAdapterName, would you like to add this?

Tools/WinMLRunner/WinMLRunner.vcxproj
jeffwick-msft: Correct, this should be removed.

ryanlai2: If I remove this, then WinMLRunner won't be able to run on builds that don't have DxCore. 

I get this error:
image

Tools/WinMLRunner/src/Run.cpp
ryanlai2: Changed to dynamicly determine size needed

Tools/WinMLRunner/src/CommandLineArgs.cpp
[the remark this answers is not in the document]
ryanlai2: Changed to DXCORE_SUPPORTED_BUILD.

Tools/WinMLRunner/src/Run.cpp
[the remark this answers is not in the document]
ryanlai2: Nevermind, I can add a back door so the user can select WARP in the future.

Tools/WinMLRunner/README.md
ryanlai2: If it's not guaranteed from execution to execution. I would propose to remove it.

ryanlai2: Removed -GpuAdapterIndex.

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -178,6 +179,119 @@ HRESULT CreateSession(LearningModelSession& session, IDirect3DDevice& winrtDevic
...
+             {
+                 throw hresult_invalid_argument(L"ERROR: No matching adapter with given adapter name: " +
+                                               adapterName);
+             }
+             size_t driverDescriptionSize;
+             THROW_IF_FAILED(spAdapter->QueryPropertySize(DXCoreProperty::DriverDescription, &driverDescriptionSize));
+             CHAR* driverDescription = new CHAR[driverDescriptionSize];
+             spAdapter->QueryProperty(DXCoreProperty::DriverDescription, driverDescriptionSize, driverDescription);
+             printf("Using adapter : %s\n", driverDescription);
+             free(driverDescription);
```
avakar, line 244: You're mixing `new[]` and `free`; this should be `delete[]`.

PR #221, opened 2019-04-22
Refactor FP16 tensor creation to prepare for client tensorization and…

pmbrown1055:
… preprocessing of images.


ryanlai2:
Looks good to me

PR #223, opened 2019-04-22
Add -onClientTensorize [Identity|ScaleMeanStdDev] functionality + uni…

pmbrown1055:
…t tests.


Tools/WinMLRunner/src/BindingUtilities.h
```diff
@@ -610,12 +675,19 @@ namespace BindingUtilities
...
+                                                   softwareBitmap);
         }
 
         auto tensorDescriptor = description.try_as<TensorFeatureDescriptor>();
         if (tensorDescriptor)
         {
+            if (args.IsImageInput())
+            {
+                softwareBitmap =
+                    LoadImageFile(tensorDescriptor, inputDataType, args.ImagePath().c_str(), args, iterationNum);
```
ryanlai2: I think it will be better to have softwareBitmap be a local variable within CreateTensor and to call LoadImageFile within there to avoid passing softwareBitmap as a parameter through CreateTensor.

ryanlai2: Also, there is an assert in LoadImageFile where inputDataType cannot be InputDataType::Tensor that needs to be taken into account.

For example, these scenarios would hit the assert statement
./winmlrunner.exe -model ./squeezenet.onnx -input C:\fish.png -tensor
./winmlrunner.exe -model ./DenseNet121_fp32.onnx -input C:\fish.png -OnClientTensorize ScaleMeanStdDev 255 123.68 116.779 103.939 -tensor

pmbrown1055: I'm ok with this... what about csv file?  Shouldn't that be treated the same way?  That's what I modelled after. :)

Is it ok to just delete this check / assert.  LoadImage is being called for both image path and tensor path... does it make sense to have any check for args when it's called?

ryanlai2: Good point! What do you think of moving the call to  ParseCSVElementStrings(args.CsvPath()); into CreateTensor<> ? Within check for args.IsCsvInput() and respectively moving the loading image into the check for args.IsImageInput() within CreateTensor<>?

ryanlai2: I think there should be a commandlineargs check in the beginning to check that -OnClientTensorize only works if image input is provided. Also another check to make sure that -OnClientTensorize and -Tensor cannot be used in the same command.

I still want this command: ./winmlrunner.exe -model ./squeezenet.onnx -input C:\fish.png -tensor to be valid though. Basically having fish be tensorized (Identity)

pmbrown1055: So, in that case, -tensor is the same as -OnClientTensorize.  What's the difference between -OnClientTensorize and -tensor?  should I just change -tensor to add some optional additional arguments?

Moving String reading and Image Loading into CreateTensor<> is fine with me.  Since that code is replicated through the template instantiation... assuming you're not worried about that?

Also, any thoughts on trying to remove the switch statement for CreateTensor, and having the default case output the error message?  I would need to think on this though... maybe punt as a future possible optimization.  There may be a way to make it really clean and still get the error checking... need to refresh my memory.

ryanlai2: If a user specifies -tensor, then the intention is to bind a tensor given either an image, csv file, or arbitrary data. In that case, -tensor and -OnClientTensorize are different in the sense that -OnClientTensorize can change <scale> <msdR> <msdB> <msdG>. -OnClientTensorize only works with input that are images right?

 I can think of a couple options, let me know what you think? : 
- -tensor takes in optional additional arguments and an error is thrown if there isn't any input image provided if optional additional arguments are specified.
- -tensor tensorizes images if one is passed in. -OnClientTensorize (Maybe should be renamed in this case) would be used to provide <scale> <msdR> <msdB> <msdG> information.  An error is thrown if -OnClientTensorize is used without the use of -Tensor.

Also, good point about the template replication. I am fine with leaving out the load image outside of the template.

ryanlai2: Can you give some more info on removing the switch statement? Why not add a default case to output error message while keeping the switch statement?

pmbrown1055: There's no reason why OnClientTensorize couldn't work with a csv input.  In fact, if you have a raw csv input, you'll get incorrect predictions with ResNet50 and DenseNet121 without input preprocessing.  I don't think it's an image only case.  Makes more sense that there would need to be no garbage data though.

I like -tensor with additional optional arguments.  Input algorithm would apply to all input data (csv or image).  Possibly PB data as well, if that's passed in -- maybe.

Keep LoadImage outside CreateTensor?

Removing the switch statement is just a thought on cleanup up the code.  No real reason to do so, just pondering... :)

pmbrown1055: Maybe if we create a data buffer for the CSV Element string, the code can just pass a pointer to the input buffer to CreateTensor... then the input is handled outside CreateTensor (and the CSV/Bitmap/PB aspects), and only preprocessing and output Tensor is done in the template?

pmbrown1055: I think the best way to clean this up would be to have all the input call generate a BYTE buffer for the input data (image and csv), then call CreateTensor with that pointer as input.

What circumstances will hit the ImageFeatureDescritor if statement, where a csv file is parsed, in CreateBindableTensor?  Looking at the code, I'm having some trouble understanding...  is it with a csv input trying to call WinML with a VideoFrame format?

ryanlai2: > I think the best way to clean this up would be to have all the input call generate a BYTE buffer for the input data (image and csv), then call CreateTensor with that pointer as input.
> 
> What circumstances will hit the ImageFeatureDescritor if statement, where a csv file is parsed, in CreateBindableTensor? Looking at the code, I'm having some trouble understanding... is it with a csv input trying to call WinML with a VideoFrame format?

I agree that a BYTE buffer for the input data and the pointer passed into CreateTensor would be the best way. :)  

To clarify your confusion, the descriptor: ImageFeatureDescriptor or TensorFeatureDescriptor is determined by the model's input. If there is an image denotation for the model's input then the descriptor will be ImageFeatureDescriptor, otherwise it will be TensorFeatureDescriptor. This is independent of whether the user passes in CSV file or Image file. The goal should be that the user is able to bind a tensor with either ImageFeatureDescriptor or TensorFeatureDescriptor using either CSV or Image file. There are restrictions for binding a tensor to ImageFeatureDescriptor because the channel (NCHW) should be either 1 or 3.

Both of these commands will hit CreateBindableTensor and ImageFeatureDescriptor is used because mnist.onnx has an Image Denotation.
image

- ./winmlrunner.exe -model ./mnist.onnx -input ./mnist_28.csv -tensor
- ./winmlrunner.exe -model ./mnist.onnx -input ./mnist_28.png -tensor

ryanlai2: > There's no reason why OnClientTensorize couldn't work with a csv input. In fact, if you have a raw csv input, you'll get incorrect predictions with ResNet50 and DenseNet121 without input preprocessing. I don't think it's an image only case. Makes more sense that there would need to be no garbage data though.
>
>I like -tensor with additional optional arguments. Input algorithm would apply to all input data (csv or image). Possibly PB data as well, if that's passed in -- maybe.
> 
I think you bring up a good point that input preprocessing should be able to work with CSV input as well. Does your design of input preprocessing apply to models that take in shapes of tensors that are not image shapes? e.g. [1,5,3,2] If not, then there should be a check to make sure that input preprocessing works with inputs that are [1,3,H,W] or [1,1,H,W].

> Keep LoadImage outside CreateTensor?
> 
Yes, I agree that keeping LoadImage outside CreateTensor is best.

> Removing the switch statement is just a thought on cleanup up the code. No real reason to do so, just pondering... :)
:)

pmbrown1055: -tensor [function] change is in progress.

Realizing that the CSV code interprets as generic strings and passes that to CreateTensor.  Interpretation of that string data is actually dependent on the output TensorKind type. So it is always assumed that the csv is in the correct input format (so string is interpreted as int32 for Int32 tensor, and as Float for Float32 or Float16 tensors).  This is a little challenging, as I can't size a BYTE buffer without knowing the type, which requires the descriptor and the number of lines in the file -- which requires reading the entire file.  Going to think on this a little more...  Suggestions welcome.

ryanlai2: Thanks so much Paul! I would like your thoughts on this approach:
- Parse the CSV file into an interpretation that is always Float32.
- Copy the resulting Float32 data into the binding buffer that casts the data into whichever type the input needs.

pmbrown1055: Can try that out.  One caveat is that the data will be interpreted into a buffer (still need to figure out how to figure out how much data needs to be read).  I _should_ be able to get that from the tensor dimensions... and fail out if the file is too small.

There may be an extra copy with this methodology... BUT, the list of strings will be eliminated, so I think overall we have a win.  This will also allow csv data to be saved as NHWC and reinterpreted as NCHW during input preprocessing...

pmbrown1055: Why is the ImageDescriptor data type and tensor type always Float32?  Is that defined that way in the API?

pmbrown1055: CSV load into Float32, then into binding, then casts data, and copies into tensor.  Would need to read binding data and interpret / preprocess to generate correct inputs, even from CSV.  Need to handle that case with as few copies as possible.

What about, interpret descriptor to determine buffer size, read csv into buffer as float32 from descriptor size definition.  Then can just loop over buffer contents and load into tensor with a single interpretation / preprocessing.  Hard part is if there are different sets of dimensions... can I expect all data to be NCHW in format?  total csv entries == batch size, channels, height, width?

ryanlai2: > Why is the ImageDescriptor data type and tensor type always Float32? Is that defined that way in the API?
May I get more clarity on your question? Target tensor descriptions must either be of type FLOAT32 or FLOAT16.

ryanlai2: >  Hard part is if there are different sets of dimensions... can I expect all data to be NCHW in format? total csv entries == batch size, channels, height, width?

Yes if CSV files are to be considered in this change, there are a bunch of edge cases where the CSV file content doesn't fit what the model expects. The edge cases need to be considered and error messages to be outputted.  Maybe it would make this PR simpler if it targeted specifically videoframe input only (PNG, JPG, etc) and not consider the CSV input.

pmbrown1055: > Why is the ImageDescriptor data type and tensor type always Float32? Is that defined that way in the API?
> 
> May I get more clarity on your question? Target tensor descriptions must either be of type FLOAT32 or FLOAT16.

Image Descriptior is always creating a Float32 input tensor (CreateTensor<TensorKind::Float32>), I was wondering if that was some kind of requirement or limitation.  I will maintain the code paths as is for now.

pmbrown1055: > Hard part is if there are different sets of dimensions... can I expect all data to be NCHW in format? total csv entries == batch size, channels, height, width?
> 
> Yes if CSV files are to be considered in this change, there are a bunch of edge cases where the CSV file content doesn't fit what the model expects. The edge cases need to be considered and error messages to be outputted.  Maybe it would make this PR simpler if videoframe input only was targeted.  (PNG, JPG, etc) and not consider the CSV input and have an issue track the work to consider CSV input in another change.

I'm not looking at VideoFrame input at all, just input images.  Once the data is in a VideoFrame I need to copy it back out to a SoftwareBitmap… so I intercept it as a SoftwareBitmap when it's first read to prevent the extra copy.

Hmm... I'm trying to put something comprehensive together... let me see if I can get it done today and have you take a look... I'm actually pretty happy with where the code is headed right now, but we'll see.

ryanlai2: Thanks so much Paul! Looking forward to your changes.

pmbrown1055:
I just pushed up the onClientTensorize branch, please take a look when you get a chance.  Lots of changes here, but csv and image are both read from disk into a buffer, which is then used to create the tensor / input preprocessed in the same loop.

One strange thing... since I added the functionality to -tensor, -bgr, causes another pass through the model and all iterations, which have a strange effect of the output tests (get double values).  Also, the user needs to input the mean std dev values in the order of the tensor channels (this enables more the 3 channel support too).

pmbrown1055:
Updated per our conversations... waiting for more feedback or merge to proceed.  I believe that everything is in place for merge.

Tools/WinMLRunner/src/CommandLineArgs.h
```diff
@@ -167,7 +193,7 @@ class CommandLineArgs
     uint32_t m_topK = 1;
     std::vector<std::pair<std::string, std::string>> m_perfFileMetadata;
 
-    void CheckNextArgument(const std::vector<std::wstring>& args, UINT i);
+    void CheckNextArgument(const std::vector<std::wstring>& args, UINT argIdx, UINT checkIdx=0);
```
ryanlai2: nit spacing: UINT checkIdx = 0;

Tools/WinMLRunner/src/BindingUtilities.h
```diff
@@ -411,74 +353,144 @@ namespace BindingUtilities
...
 
         com_ptr<ITensorNative> spTensorValueNative;
         tensorValue.as(spTensorValueNative);
 
-        PointerType* actualData;
+        WriteType* actualData;
         uint32_t actualSizeInBytes;
-        spTensorValueNative->GetBuffer(reinterpret_cast<BYTE**>(&actualData),
-                                                                &actualSizeInBytes);
+        spTensorValueNative->GetBuffer(reinterpret_cast<BYTE**>(&actualData), &actualSizeInBytes);
```
ryanlai2: Can a "THROW_IF_FAILED()" be wrapped around this?

PR #229, opened 2019-05-06
User/orilevari/update dashboard readme

orilevari:
Updates the dashboard readme to include an image of the output debugging convolution with png output type.
Updates readme to point to newer tag (which is yet to be created since I will target the tag at this pull request merge).


Tools/WinMLDashboard/README.md
```diff
@@ -75,13 +75,16 @@ To debug your model follow these steps:
...
+
+*Intermediate PNG convolutional output from an image of a [kitten](../../SharedContent/media/kitten_224.png)*
 
 
 ## Install
 
 You can install an early preview version of WinML Dashboard from here.
 
-[Download Preview Version](https://github.com/Microsoft/Windows-Machine-Learning/releases/tag/v0.5.1)
+[Download Preview Version](https://github.com/Microsoft/Windows-Machine-Learning/releases/tag/v0.6.1)
```
zhangxiang1993: Did you talk with Kumar about the version number?

orilevari: Yep, this is the version number we agreed on.

Tools/WinMLDashboard/README.md
```diff
@@ -75,13 +75,16 @@ To debug your model follow these steps:
 
 <img src="./public/ConfigureDebug.png" width=800/>
 <img src="./public/DebugView.png" width=800/>
+<img src="./public/DebugPngs.png" width=800/>
```
zhangxiang1993: Can you add like one-line description before showing the debug output?

orilevari: There is a one line description under the image. Would you prefer that to be above the image?

zhangxiang1993: I think so. Separate the output image and the image of how to use above with words might be a better idea.

PR #231, opened 2019-05-07
update tensorflow package for python environment

orilevari:
The previous version of tensorflow had a security vulnerability reported via GitHub so this pull request updates the package with a bug fix patch. I tested to make sure the tensorflow converter still works and the converted model can be run using the dashboard.

Here is the original alert message:
alert


zhangxiang1993:
Can you update the status to the build pipeline moved to OS repo in the [readme](https://github.com/microsoft/Windows-Machine-Learning/blob/master/README.md)?

orilevari:
Status badge has been updated

PR #232, opened 2019-05-07
update all version information for dashboard to v0.6.1

orilevari:
updates version info to 0.6.1.
The download link in the root readme previously was pointing to the wrong url so now that is fixed as well.


orilevari:
I'm not sure why the ci build is pointing to the out of date mscodehub pipeline, seems like the up to date pipeline is passing though.

PR #237, opened 2019-05-29
This change allows user to specify folder of images to bind and evaluate

ryanlai2:
New flag: -InputImageFolder

 .\WinMLRunner.exe -model .\SqueezeNet.onnx -inputimagefolder .\images\

Also this change allows relative paths for specifying path to all inputs

### DIFF
--- a/Tools/WinMLRunner/src/CommandLineArgs.cpp
+++ b/Tools/WinMLRunner/src/CommandLineArgs.cpp
@@ -36,6 +36,7 @@ void CommandLineArgs::PrintUsage()
               << std::endl;
     std::cout << "  -Iterations : # times perf measurements will be run/averaged. (maximum: 1024 times)" << std::endl;
     std::cout << "  -Input <fully qualified path> : binds image or CSV to model" << std::endl;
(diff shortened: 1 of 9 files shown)


Tools/WinMLRunner/src/CommandLineArgs.cpp
```diff
@@ -420,4 +447,8 @@ void CommandLineArgs::CheckForInvalidArguments()
     {
         throw hresult_invalid_argument(L"Cannot save tensor output if no input data is provided!");
     }
+    if (m_imagePaths.size() > 1 && IsSaveTensor())
+    {
+        throw hresult_not_implemented(L"Saving tensor output for multiple images isn't implemented yet.");
```
orilevari: I'm not sure if you should say implemented _yet_. It in a way commits you to the public to implement this in the future, and I don't think we should be putting those sorts of commitments in code comments. What do you think?

ryanlai2: I completely agree with your points.  Let me remove " _yet_ "

PR #239, opened 2019-05-31
User/brian/steamfromresource

martinb35:
new sample app that has an embedded model in the app's resource file and then uses load from stream to create the learningModel.


orilevari:
Can you add a simple validation test in the public samples release pipeline task group?

orilevari:
The readme included for this sample doesn't seem to be explaining the winml sample, but instead explains c++/winrt development in general. Can you include a readme that explains the motivation behind this sample?

PR #248, opened 2019-06-19
Feature/std dev

laks0209:
Added std dev & corrected -SavePerIteration … 
* Standard deviation values are printed to screen when -perf all command
line is given. They are also written to file if -perfOutput is given
* Corrected - SavePerIteration issue (now outputs eval & bind times for
all iterations)

### DIFF
--- a/Tools/WinMLRunner/src/OutputHelper.h
+++ b/Tools/WinMLRunner/src/OutputHelper.h
@@ -202,11 +202,13 @@ public:
         double createSessionTime = profiler[CREATE_SESSION].GetAverage(CounterType::TIMER);
 
         double averageBindTime = profiler[BIND_VALUE].GetAverage(CounterType::TIMER);
+        double stdevBindTime = profiler[BIND_VALUE].GetStdev(CounterType::TIMER);
         double minBindTime = profiler[BIND_VALUE].GetMin(CounterType::TIMER);
         double maxBindTime = profiler[BIND_VALUE].GetMax(CounterType::TIMER);
         double firstBindTime = profiler[BIND_VALUE_FIRST_RUN].GetAverage(CounterType::TIMER);
 
         double averageEvalTime = profiler[EVAL_MODEL].GetAverage(CounterType::TIMER);
+        double stdevEvalTime = profiler[EVAL_MODEL].GetStdev(CounterType::TIMER);
         double minEvalTime = profiler[EVAL_MODEL].GetMin(CounterType::TIMER);
         double maxEvalTime = profiler[EVAL_MODEL].GetMax(CounterType::TIMER);
         double firstEvalTime = profiler[EVAL_MODEL_FIRST_RUN].GetAverage(CounterType::TIMER);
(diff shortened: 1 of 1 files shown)


Tools/WinMLRunner/src/OutputHelper.h
```diff
@@ -518,16 +540,18 @@ class OutputHelper
...
     }
     void SaveEvalPerformance(Profiler<WINML_MODEL_TEST_PERF>& profiler, uint32_t iterNum)
     {
-        m_clockEvalTimes[iterNum] = profiler[EVAL_MODEL].GetClockTime();
-        m_CPUWorkingDiff[iterNum] = profiler[EVAL_MODEL].GetCpuWorkingDiff();
-        m_CPUWorkingStart[iterNum] = profiler[EVAL_MODEL].GetCpuWorkingStart();
-        m_GPUSharedDiff[iterNum] = profiler[EVAL_MODEL].GetGpuSharedDiff();
-        m_GPUSharedStart[iterNum] = profiler[EVAL_MODEL].GetGpuSharedStart();
-        m_GPUDedicatedDiff[iterNum] = profiler[EVAL_MODEL].GetGpuDedicatedDiff();
+        iterationNum = (iterNum == 0) ? EVAL_MODEL_FIRST_RUN : EVAL_MODEL;
```
ryanlai2: Any reason why iterationNum is being used like this here as an index to access the performance arrays (m_clockEvalTimes, m_CPUWorkingDiff, m_CPUWorkingStart, etc, etc)? It will always either be EVAL_MODEL_FIRST_RUN (5) or EVAL_MODEL (6)..

Tools/WinMLRunner/src/OutputHelper.h
```diff
@@ -518,16 +540,18 @@ class OutputHelper
...
+        m_clockEvalTimes[iterationNum] = profiler[EVAL_MODEL_FIRST_RUN].GetClockTime();
+        m_CPUWorkingDiff[iterationNum] = profiler[EVAL_MODEL_FIRST_RUN].GetCpuWorkingDiff();
+        m_CPUWorkingStart[iterationNum] = profiler[EVAL_MODEL_FIRST_RUN].GetCpuWorkingStart();
```
ryanlai2: profiler[EVAL_MODEL_FIRST_RUN] will always return the performance of the first evaluate. Is this intended?

Tools/WinMLRunner/src/OutputHelper.h
```diff
@@ -986,6 +1042,8 @@ class OutputHelper
     std::vector<std::string> m_outputResult;
     std::vector<int> m_outputTensorHash;
 
+    int iterationNum;
```
ryanlai2: What is this iterationNum variable used for?

laks0209:
The idea was to use iterationNum as an index to the profiler so that for when Iteration number=0, profiler[EVAL_MODEL_FIRST_RUN] is used and other times profiler[EVAL_MODEL] is used. Due to several resets in the head, i pushed in a wrong commit. I have made sure everything is correct now. Please take a look into the recent commit.

PR #256, opened 2019-07-11
User/orilevari/winmlrunner dxcore latest

orilevari:
update dxcore usage to latest api surface (public as of build 18936)

remove enable experimental features call


Tools/WinMLRunner/src/Run.cpp
```diff
@@ -290,19 +290,15 @@ HRESULT CreateSession(LearningModelSession& session, IDirect3DDevice& winrtDevic
                      // If DXGI factory creation was successful then get the IDXGIAdapter from the LUID acquired from
                      // the selectedAdapter
                      std::cout << "Using DXGI for adapter creation.." << std::endl;
+                     size_t luidSize;
```
martinb35: initialize luidSize to 0?

orilevari: You know, looking at this I'm realizing I should really be using the templated version of GetProperty that doesn't accept a size parameter for properties that aren't variable length.

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -291,18 +291,12 @@ HRESULT CreateSession(LearningModelSession& session, IDirect3DDevice& winrtDevic
...
                      THROW_IF_FAILED(dxgiFactory4->EnumAdapterByLuid(adapterLuid, __uuidof(IDXGIAdapter),
                                                                      spDxgiAdapter.put_void()));
                      pAdapter = spDxgiAdapter.get();
                  }
              }
-             else
-             {
-                 // Need to enable experimental features to create D3D12 Device with adapter that has compute only
-                 // capabilities.
-                 THROW_IF_FAILED(D3D12EnableExperimentalFeatures(1, &D3D12ComputeOnlyDevices, nullptr, 0));
```
walrusmcd: >D3D12EnableExperimentalFeatures [](start = 33, length = 31)

why did we remove this one?

walrusmcd:
> #ifdef DXCORE_SUPPORTED_BUILD

For future revs, can we change this to not be a compiler #ifdef, but instead a runtime OS check?  that way a single binary can support both DXGI and DXCORE, right?    with this ifdef I have to rebuild seperate binaries?

---
Refers to: Tools/WinMLRunner/src/Run.cpp:205 in 7573da4. [](commit_id = 7573da4a9ea8b00e34efdddc826f493448cfe49e, deletion_comment = False)

Tools/WinMLRunner/src/Run.cpp
orilevari: "Simon has done the work, It is no longer needed and we can use MCDM devices without enabling Developer Mode.
D3D12EnableExperimentalFeatures() fails if Developer Mode is not enabled and that causes WinMLRunner.exe to exit."

PR #261, opened 2019-07-24
allow specifying an adapter name also when using old DXGI

deischi:
I see no need to use DXCORE to specify an adapter name.

### DIFF
--- a/Tools/WinMLRunner/src/CommandLineArgs.cpp
+++ b/Tools/WinMLRunner/src/CommandLineArgs.cpp
@@ -21,10 +21,7 @@ void CommandLineArgs::PrintUsage()
     std::cout << "  -GPU : run model on default GPU" << std::endl;
     std::cout << "  -GPUHighPerformance : run model on GPU with highest performance" << std::endl;
     std::cout << "  -GPUMinPower : run model on GPU with the least power" << std::endl;
-#ifdef DXCORE_SUPPORTED_BUILD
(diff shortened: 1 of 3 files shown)


ryanlai2:
Hi @deischi, thank you for your contribution! It's correct that specifying adapter name shouldn't be constrained to DxCore.

I have a few suggested changes:
- There is already a conditional to handle if "-GPUAdapterName" flag is used. Can your code to choose adapter be added inside here?: 
https://github.com/microsoft/Windows-Machine-Learning/blob/ed0b176a8b4a273ae8ce124b507ba833fc30ecb6/Tools/WinMLRunner/src/Run.cpp#L210

- Within the conditional of (WinMLDeviceKind isn't CPU and adaptername isn't empty). One flow can be like this:
  - Try to use DXGI to enumerate adapters
  - Fall back to using DxCore if DXGI doesn't work / isn't available.
  - #ifdef check for DXCORE_SUPPORTED_BUILD will need to be around the DxCore code.
  - Create Session with chosen adapter.

- Check for first matching substring for adapter name.

- Spaces instead of tabs for spacing.

Thanks! Let me know if you have any questions.

deischi:
I have tried to add my changes as local and small as possible to quickly get a result.

You are right, having the adapter selection only once would be better.
Please understand that improving the WinMlRunner is not one of the top priorities, so it will definitely take some time until come to that.
Also I have no access to DXCore, any changes I will do related to that will be done blindly.

PR #268, opened 2019-08-22
User/orilevari/dashboard free dim input

orilevari:
Source copy newest commit of WinMLRunner to be consumed by WinMLDashboard and modify required File loading paths to register the custom debug operator. This change was originally motivated by the inability to evaluate models with free input dimensions due to using old WinMLRunner sources.


orilevari:
This pull request does another source copy of WinMLRunner for the purposes of having a flavor that can consume the Debug operator.

zhangxiang1993:
Did we decide to copy WinmlRunner source code to WinmlDashboard? Could we do a git submodule?

orilevari:
As we discussed in the sync up last week, this pr unblocks the free dim bug and continues the previous solution of source copying. Investigation into a potential better overall solution (i.e flag in WinMLRunner or git submodule) should be decoupled from fixing this bug.

zhangxiang1993:
I'll approve this for the current solution as discussed, and please file a task/bug for Investigation into a potential better overall solution

PR #280, opened 2019-10-08
Feature/garbage data range

pmbrown1055:
Adds the ability to limit the garbage data range.  Useful if input data is an index into another buffer or memory.


ryanlai2:
Looks good to me!

PR #282, opened 2019-10-15
Update Squeezenet c++ and C# sample to be explicit with colorspace management of GetSoftwareBitmapAsync

ryanlai2:


Samples/SqueezeNetObjectDetection/Desktop/cpp/main.cpp
```diff
@@ -153,7 +158,32 @@ void LoadLabels()
...
 }
 
-VideoFrame LoadImageFile(hstring filePath)
+ColorManagementMode GetColorManagementMode(const LearningModel& model)
+{
+    // Get model color space gamma
+    hstring gammaSpace = L"";
+    try
+    {
+        gammaSpace = model.Metadata().Lookup(L"Image.ColorSpaceGamma");
```
tiagoshibata: `Keys and values are case insenstive` according to https://github.com/onnx/onnx/blob/master/docs/MetadataProps.md , so  e.g.`image.colorspacegamma` is valid too. The same goes for e.g. `sRGB` instead of `SRGB`.

Samples/SqueezeNetObjectDetection/Desktop/cpp/main.cpp
```diff
@@ -153,7 +158,32 @@ void LoadLabels()
...
+    if (gammaSpace == L"" || gammaSpace == L"SRGB")
+    {
+        return ColorManagementMode::ColorManageToSRgb;
+    }
+    else
+    {
+        // Due diligence should be done to make sure that the input image is within the model's colorspace. There are multiple non-sRGB color spaces.
+        printf("    Model metadata indicates that color gamma space is : %ws. Will not manage color space...\n", gammaSpace.c_str());
+        return ColorManagementMode::DoNotColorManage;
+    }
```
tiagoshibata: Removing the else is equivalent and shorter:

```
    if (gammaSpace == L"" || gammaSpace == L"SRGB")
    {
        return ColorManagementMode::ColorManageToSRgb;
    }

    // Due diligence should be done to make sure that the input image is within the model's colorspace. There are multiple non-sRGB color spaces.
    printf("    Model metadata indicates that color gamma space is : %ws. Will not manage color space...\n", gammaSpace.c_str());
    return ColorManagementMode::DoNotColorManage;
```

Samples/SqueezeNetObjectDetection/Desktop/cpp/main.cpp
ryanlai2: Sounds good! Good catch :)

Samples/SqueezeNetObjectDetection/NETCore/cs/Program.cs
```diff
@@ -117,17 +120,53 @@ private static T AsyncHelper<T> (IAsyncOperation<T> operation)
             return operation.GetResults();
         }
 
-        private static ImageFeatureValue LoadImageFile()
+        private static ImageFeatureValue LoadImageFile(ColorManagementMode colorManagementMode)
         {
             StorageFile imageFile = AsyncHelper(StorageFile.GetFileFromPathAsync(_imagePath));
```
orilevari: Can you add a try catch on loading the file with the fully qualified path error message

Samples/SqueezeNetObjectDetection/Desktop/cpp/main.cpp
```diff
@@ -163,16 +189,33 @@ VideoFrame LoadImageFile(hstring filePath)
...
+
         // load a videoframe from it
         VideoFrame inputImage = VideoFrame::CreateWithSoftwareBitmap(softwareBitmap);
         // all done
         return inputImage;
     }
     catch (...)
     {
-        printf("failed to load the image file, make sure you are using fully qualified paths\r\n");
+        printf("    Failed to load the image file, make sure you are using fully qualified paths\r\n");
```
orilevari: Can you put this catch closer to the actual API that it is related to?

Samples/SqueezeNetObjectDetection/NETCore/cs/Program.cs
ryanlai2: Yep yep!

Samples/SqueezeNetObjectDetection/Desktop/cpp/main.cpp
ryanlai2: Yep yep!

Samples/SqueezeNetObjectDetection/NETCore/cs/Program.cs
```diff
@@ -117,17 +120,53 @@ private static T AsyncHelper<T> (IAsyncOperation<T> operation)
...
+            if (!doesModelContainGammaSpaceMetadata)
+            {
+                Console.WriteLine("    Model does not have color space gamma information. Will color manage to sRGB by default...");
+            }
+            if (!doesModelContainGammaSpaceMetadata || gammaSpace.Equals("SRGB", StringComparison.CurrentCultureIgnoreCase))
+            {
+                return ColorManagementMode.ColorManageToSRgb;
+            }
+            // Due diligence should be done to make sure that the input image is within the model's colorspace. There are multiple non-sRGB color spaces.
+            Console.WriteLine("    Model metadata indicates that color gamma space is : {0}. Will not manage color space...", gammaSpace);
```
ZuomingShi: Suggest "Will not manage color space to sRGB..." instead of "Will not manage color space..."

Samples/SqueezeNetObjectDetection/NETCore/cs/Program.cs
```diff
@@ -117,17 +120,53 @@ private static T AsyncHelper<T> (IAsyncOperation<T> operation)
...
             softwareBitmap = SoftwareBitmap.Convert(softwareBitmap, BitmapPixelFormat.Bgra8, BitmapAlphaMode.Premultiplied);
             VideoFrame inputImage = VideoFrame.CreateWithSoftwareBitmap(softwareBitmap);
             return ImageFeatureValue.CreateFromVideoFrame(inputImage);
         }
 
+        private static ColorManagementMode GetColorManagementMode()
+        {
+            // Get model color space gamma
+            string gammaSpace = "";
+            bool doesModelContainGammaSpaceMetadata = _model.Metadata.TryGetValue("Image.ColorSpaceGamma", out gammaSpace);
```
ZuomingShi: Out of curiosity is there documentation around the meaning of "Image.ColorSpaceGamma"? I couldn't find it online after a quick search.

ryanlai2: Yep! https://github.com/onnx/onnx/blob/master/docs/MetadataProps.md

Samples/SqueezeNetObjectDetection/Desktop/cpp/main.cpp
tiagoshibata: Is this resolved? Is the `Lookup` function already case-insensitive?

ryanlai2: model.Metadata() is a IMapView interface so it'll be up to WinML to handle the case-insensitive key lookup :)

ryanlai2: "Image.ColorSpaceGamma" should return the same value as "image.ColorSpaceGamma"

tiagoshibata: Okay then :)

PR #283, opened 2019-10-15
Update WinMLRunner to be explicit with colorspace management of GetSoftwareBitmapAsync

ryanlai2:
Managing color space gamma only is applicable when an image is passed in as input.


Tools/WinMLRunner/src/BindingUtilities.h
```diff
@@ -247,21 +272,23 @@ namespace BindingUtilities
...
-                return decoder.GetSoftwareBitmapAsync(format, BitmapAlphaMode::Ignore).get();
+                return decoder.GetSoftwareBitmapAsync(format, decoder.BitmapAlphaMode(), BitmapTransform(),
+                                                      ExifOrientationMode::RespectExifOrientation, colorManagementMode).get();
             }
         }
         catch (...)
         {
             std::wcout << L"BindingUtilities: could not open image file (" << std::wstring(filePath) << L"), "
-                       << L"make sure you are using fully qualified paths." << std::endl;
+                       << L"make sure you are using fully qualified paths. Also please make sure that input image is within the model's colorspace. " << std::endl;
```
orilevari: try to separate out the two different error conditions in separate catch blocks

PR #297, opened 2020-02-29
Re-add lock file, fix build breaks

tiagoshibata:
Builds were failing after the lockfile was deleted by accident and some transitive dependencies were updated. It was deleted a long time ago (before WinML Dashboard was moved to this repo - it's available in the internal WindowsAI repo history only).

babel-minify is failing to minify Netron with a stack depth limit reached. I don't know if it's a bug in babel-minify with some infinite recursion, or the Netron source is too large for it to handle. Since the dashboard is low priority and not in my taskboard, I simply disabled minifying Netron for now, which adds a few hundred kilobytes to the build. We can investigate this later.

Since I was adjusting the packages and fixing build breaks I also updated the packages in package.json, which were quite outdated.


Tools/WinMLDashboard/package.json
```diff
@@ -33,42 +33,42 @@
...
+    "fs-extra": "^8.1.0",
     "md5": "^2.2.1",
-    "mocha": "^5.2.0",
+    "mocha": "^7.1.0",
     "ncp": "^2.0.0",
     "npm-font-open-sans": "^1.1.0",
     "office-ui-fabric-react": "^6.40.0",
     "react": "^16.5.2",
-    "react-app-rewired": "^1.5.2",
+    "react-app-rewired": "^2.1.5",
```
zhangxiang1993: Please use the lower version of react-app-rewired  because of this [https://github.com/timarney/react-app-rewired/issues/358](https://github.com/timarney/react-app-rewired/issues/358)

Tools/WinMLDashboard/src/view/run/View.tsx
```diff
@@ -322,9 +322,13 @@ class RunView extends React.Component<IComponentProperties, IComponentState> {
             });
     }
 
-    private logError = (error: string | Error) => {
-        const message = typeof error === 'string' ? error : (`${error.stack ? `${error.stack}: ` : ''}${error.message}`);
-        log.error(message)
+    private logError = (error: string | Error | Error[]) => {
+        const toString = (error: string | Error) => typeof error === 'string' ? error : (`${error.stack ? `${error.stack}: ` : ''}${error.message}`);
```
zhangxiang1993: TSLint would cause shadowing variable here.

PR #298, opened 2020-03-03
update winmldashboard to use latest winmlrunner and winmltools

zhangxiang1993:


Tools/WinMLDashboard/README.md
```diff
@@ -1,6 +1,6 @@
 # WinML Dashboard (preview)
 
-WinML Dashboard is a tool for **viewing**, **editing**, **converting**, and **validating** machine learning models for [Windows ML](https://docs.microsoft.com/en-us/windows/ai/) inference engine. The engine is built into Windows 10 and evaluates trained models locally on Windows devices using hardware optimizations for CPU and GPU to enable high performance inferences. [(Download Preview Version of the tool)](https://github.com/Microsoft/Windows-Machine-Learning/releases/tag/v0.6.1)   
+WinML Dashboard is a tool for **viewing**, **editing**, **converting**, and **validating** machine learning models for [Windows ML](https://docs.microsoft.com/en-us/windows/ai/) inference engine. The engine is built into Windows 10 and evaluates trained models locally on Windows devices using hardware optimizations for CPU and GPU to enable high performance inferences. [(Download Preview Version of the tool)](https://github.com/Microsoft/Windows-Machine-Learning/releases/tag/v0.7.0)
```
martinb35: when will we be at version 1.0? 0.7.0 makes it seem like it's still preview. is it?

zhangxiang1993: Hmm, Sounds true. But this app is not codesigned, so I think stay as preview should be better.

---
In reply to: [386739110](https://github.com/microsoft/Windows-Machine-Learning/pull/298#discussion_r386739110) [](ancestors = 386739110)

PR #299, opened 2020-03-04
User/xianz/winmlrunner fail

zhangxiang1993:


Tools/WinMLRunner/WinMLRunner.sln
```diff
@@ -59,14 +59,12 @@ Global
 		{C3BCBEA1-90E6-426F-88AC-64C274BCEF45}.Release|x64.Build.0 = Release|x64
 		{C3BCBEA1-90E6-426F-88AC-64C274BCEF45}.Release|x86.ActiveCfg = Release|Win32
 		{C3BCBEA1-90E6-426F-88AC-64C274BCEF45}.Release|x86.Build.0 = Release|Win32
-		{C174D45D-C189-475B-B1A7-494939EE7491}.Debug|ARM64.ActiveCfg = Debug|ARM64
```
ryanlai2: debug arm64 should build debug arm64 and not debug win32

zhangxiang1993: Reverted the change

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
```diff
@@ -887,7 +865,7 @@ namespace WinMLRunnerTest
         TEST_METHOD(LoadModelFailModelNotFound)
         {
             const std::wstring command = BuildCommand({ EXE_PATH, L"-model", L"invalid_model_name" });
-            Assert::AreEqual(HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND), RunProc((wchar_t *)command.c_str()));
+            Assert::AreEqual(HRESULT_FROM_WIN32(E_FAIL), RunProc((wchar_t *)command.c_str()));
         }
```
martinb35: this seems like a bug in winml - file missing should be file not found, not e_fail. can we log the bug in winml too?

PR #303, opened 2020-03-23
update appx certificates for samples

orilevari:
All our certificates for our samples expired so this PR regenerates them.


Samples/Emoji8/UWP/cs/Emoji8/Emoji8.csproj
```diff
@@ -17,14 +17,16 @@
     <FileAlignment>512</FileAlignment>
     <ProjectTypeGuids>{A5A43C5B-DE2A-4C0C-9213-0A381AF9435A};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
     <WindowsXamlEnableOverview>true</WindowsXamlEnableOverview>
-    <PackageCertificateKeyFile>Emoji8_TemporaryKey.pfx</PackageCertificateKeyFile>
+    <PackageCertificateKeyFile>
```
martinb35: the certificates used to have names - do they not need those anymore?

Samples/SqueezeNetObjectDetection/NETCore/cs/SqueezeNetObjectDetectionNC.csproj
```diff
@@ -1,66 +1,57 @@
...
+<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <ApplicationIcon />
     <Win32Resource />
     <LangVersion>7.3</LangVersion>
     <Platforms>AnyCPU;x86;x64</Platforms>
+    <FileUpgradeFlags>40</FileUpgradeFlags>
+    <UpgradeBackupLocation>D:\repos\Windows-Machine-Learning\Samples\SqueezeNetObjectDetection\Backup\NETCore\cs\</UpgradeBackupLocation>
```
martinb35: do you want this change? looks like it's pointing at a local folder

orilevari: no, good catch

Samples/Emoji8/UWP/cs/Emoji8/Emoji8.csproj
orilevari: This reformatting is how visual studio updated the project files automatically. It seems like it is now inferring the pfx file just by it being included in the output directory. My thought is to not mess with how visual studio is doing it, and that in more recent visual studio versions they must have changed it. I will run a pipeline once before merging to ensure it still works remotely though.

PR #309, opened 2020-04-09
User/xianz/update sample 2

zhangxiang1993:
This PR is to 
1. standardize the output directory for all samples, so that we could set the output directory in one format in the pipeline.
2. Update certificates for all cs samples which all have expired certificates that failed the build.
3. Modified [Sample pipeline](https://microsoft.visualstudio.com/WindowsAI/_build?definitionId=39302&_a=summary) to have tests for several samples, including:
 - AdapterSelection
 - CustomOperator
 - BatchSupport
 - CustomTensorization
 - StreamFromResouce

4. Temporary disabled build of SqueezeNetObjectDetection Sample in pipeline, because of the failure of building SqueezeNetObjectDetetion NC sample.


azure-pipelines-samples.yml
```diff
@@ -256,4 +560,4 @@ steps:
     inputs:
       pathtoPublish: '$(Build.ArtifactStagingDirectory)'
       artifactName: WinMLPublicSamples-RS5
```
orilevari, line 562: Let's just change the artifactName to WinMLPublicSamples, since we aren't building against different SDKs anymore.
Since we are moving the tests into this build pipeline, there's no need to worry about changing the artifact name in the test pipeline to match.

azure-pipelines-samples.yml
```diff
@@ -134,18 +357,19 @@ steps:
...
-      solution: Samples/SqueezeNetObjectDetection/squeezenetobjectdetection.sln
-      vsVersion: 15.0
-      msbuildArgs: '-v:diag /p:OutDir=$(System.DefaultWorkingDirectory)\bin\$(BuildPlatform)\$(BuildConfiguration)\SqueezeNetObjectDetection\'
-      platform: '$(BuildPlatform)'
-      configuration: '$(BuildConfiguration)'
-      clean: true
-      msbuildArchitecture: x64
-      createLogFile: true
-    condition: succeededOrFailed()
+  # TODO: Bug 25927815: SqueezeNetObjectDetection NetCore Sample is failing to be built
```
orilevari, line 360: Make sure to add the SqueezeNetObjectDetection test back in when this get's uncommented. Maybe it would be best to add it in comments now?

azure-pipelines-samples.yml
```diff
@@ -107,6 +257,79 @@ steps:
...
+      targetType: 'inline'
+      script: |
+        $kitten = Resolve-Path ..\..\..\..\SharedContent\media\kitten_224.png
+        .\custom-operator-sample.exe relu
+        # registering new ops will only work for public sample on RS5 due to opset choice in code
+        $osinfo = Get-CimInstance Win32_OperatingSystem
+        $osinfo | Format-List *
+        $osversion = $osinfo.Version 
+        Write-Host "OS Version: $osversion"
+        if ($osversion -eq '10.0.17763') {
```
orilevari, line 287: We should really update the custom op sample to run correctly on 19H1 binaries. Or ideally be independent of binary (although I'm not sure this is possible). This may be out of scope of this PR but we should file some work for it.

PR #310, opened 2020-04-16
User/xianz/set target sdk

zhangxiang1993:


Samples/CustomTensorization/CustomTensorization/CustomTensorization.vcxproj
```diff
@@ -8,7 +8,7 @@
     <ProjectGuid>{4420ecd4-ad04-4183-9879-7d52039e5422}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
     <RootNamespace>CustomTensorization</RootNamespace>
-    <WindowsTargetPlatformVersion Condition=" '$(WindowsTargetPlatformVersion)' == '' ">10.0.17763.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion Condition=" '$(WindowsTargetPlatformVersion)' == '' ">10.0.18362.0</WindowsTargetPlatformVersion>
     <WindowsTargetPlatformMinVersion>10.0.17134.0</WindowsTargetPlatformMinVersion>
```
orilevari, line 12: I wonder why some of the samples set min version to be 17134 instead of 17763?

Samples/CustomOperator/desktop/cpp/custom-operator-sample.vcxproj
```diff
@@ -9,7 +9,7 @@
     <ProjectGuid>{2bf804d4-daa2-42be-9f21-0e94f021ef53}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
     <RootNamespace>SqueezeNetObjectDetection</RootNamespace>
-    <WindowsTargetPlatformVersion Condition=" '$(WindowsTargetPlatformVersion)' == '' ">10.0.17763.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion Condition=" '$(WindowsTargetPlatformVersion)' == '' ">10.0.18362.0</WindowsTargetPlatformVersion>
```
orilevari, line 12: Can you verify that the custom operator sample still works for registering new operators why building against 19H1 SDK? I believe that the functionality may depend on a RS5 SDK version currently, in which case the code should be modified.

azure-pipelines-samples.yml
```diff
@@ -6,6 +6,7 @@ resources:
 variables:
   SamplesBin: SamplesBin
   WINDOWS_WINMD: C:\Program Files (x86)\Windows Kits\10\UnionMetadata\10.0.18362.0\Windows.winmd
+  WindowsTargetPlatformVersion: 10.0.18362.0
```
orilevari, line 9: Why do we need to override the target platform version in our build pipeline? Shouldn't we be building it in the way that our samples are configured externally?

zhangxiang1993: This is to give a way where we can configure the target platform for all samples in one place. In the future, we don't need to update the sample, but change the sdk version here.

---
In reply to: [409810612](https://github.com/microsoft/Windows-Machine-Learning/pull/310#discussion_r409810612) [](ancestors = 409810612)

Samples/CustomTensorization/CustomTensorization/CustomTensorization.vcxproj
zhangxiang1993: This Sample works for OS low to 17134, the mininum version is the minumum requirement, I think it should be okay.

---
In reply to: [409808882](https://github.com/microsoft/Windows-Machine-Learning/pull/310#discussion_r409808882) [](ancestors = 409808882)

azure-pipelines-samples.yml
orilevari: But people that are are using and building the samples on their own will still need the windows target platform version attribute to be set correctly in the project file right? So we'll still need to update all the project files for them.

Samples/CustomTensorization/CustomTensorization/CustomTensorization.vcxproj
orilevari: 17134 if RS4, and wasn't winml release under a different preview namespace at that time? If so that would break building these samples.

Samples/CustomOperator/desktop/cpp/custom-operator-sample.vcxproj
zhangxiang1993: Such functionality failure will be caught by Sample Tests pipeline. And the pipeline will tell us whether code should be modified. What do you think?

---
In reply to: [409809598](https://github.com/microsoft/Windows-Machine-Learning/pull/310#discussion_r409809598) [](ancestors = 409809598)

orilevari: I believe the pipeline currently will just skip that part of the custom op code if the agent isn't running RS5. So if you are going to rely on the samples pipeline then, just make sure it runs on an RS5 agent, but with a 19H1 SDK. 

Ideally we should figure out a way to make the sample version independent if it is an issue.

Samples/CustomTensorization/CustomTensorization/CustomTensorization.vcxproj
zhangxiang1993: Hmm, I believe when these sample were created, they should be tested to build against RS4, I don't have 17134 installed. But I think it's also okay to change it to 17763. I'll do that.

---
In reply to: [409814617](https://github.com/microsoft/Windows-Machine-Learning/pull/310#discussion_r409814617) [](ancestors = 409814617)

Samples/CustomOperator/desktop/cpp/custom-operator-sample.vcxproj
orilevari: I think the dependency might just be at runtime though, so perhaps the SDK doesn't matter.

Samples/CustomTensorization/CustomTensorization/CustomTensorization.vcxproj
orilevari: Okay sounds good. Someone may have updated the namespace but forgot to update the min version since they had a more recent SDK installed.

orilevari: Yeah, for example back in RS4 it was a LearningModelPreview instead of LearningModel

azure-pipelines-samples.yml
orilevari: oh I understand. Sounds good!

PR #313, opened 2020-05-22
build squeezenetNC sample in pipeline

zhangxiang1993:


azure-pipelines-samples.yml
```diff
@@ -135,21 +135,6 @@ steps:
       createLogFile: true
     condition: succeededOrFailed()
 
-  # Bug 25927815: SqueezeNetObjectDetection NetCore Sample is failing to be built
-  # Also the test should be added when the code below is uncommented
```
ryanlai2, line 139: How does removing this task entirely help build the squeezenetNC  sample?

zhangxiang1993: SquezenetNC sample was disabled to be built. now we bring it back. The build of squeezenet was commented out.

smk2007: It looks like you are deleting the whole task? Is the build being done somewhere else?

Issue #324, opened 2020-07-13
WinMLRunner - remove 5D block on GPU (bug)


fdwr:
## Current behavior
WinMLRunner rejects 5D tensors, with the message "Input feature input shape is too large. GPU path only accepts tensor dimensions <= 4 : 5", even though DirectML supports 5D tensors (and will likely support more than that too).

https://github.com/microsoft/Windows-Machine-Learning/blob/master/Tools/WinMLRunner/src/Run.cpp#L296

## Expected behavior
WinMLRunner should just let any dimension count through - let the API decide if it is support or not.

## Minimal reproduction of the problem with instructions
Any model with 5D bindings.

`WinMLRunner.exe -gpu -model unet3d.channels_first.keras2onnx.onnx`

(source is models.zip from here: https://github.com/microsoft/Windows-Machine-Learning/issues/215)

## Environment

Windows Build Number: 18363.900

App min and target version: Current master, commit 8ca3e52 on 2019 Oct 15

OS Version (Server, IoT Core, Desktop, etc): Desktop

Graphics Driver version: <any>

DxDiag: <NA>

WinMLTools specific:
	- Source training framework: (e.g. CoreML, Scikit-learn, …)
        - WinMLTools version

Visual Studio 
- [ ] 2017 (version: NA)
- [ ] 2017 Preview (version: NA)

ryanlai2:
Released WinMLRunner with bug fix: https://github.com/microsoft/Windows-Machine-Learning/releases/tag/1.2.1.2

PR #325, opened 2020-07-13
Remove artifical limitation on 5D tensors for GPU 

ryanlai2:
Fixes this bug: https://github.com/microsoft/Windows-Machine-Learning/issues/324

WinMLRunner rejects 5D tensors, with the message "Input feature input shape is too large. GPU path only accepts tensor dimensions <= 4 : 5", even though DirectML supports 5D tensors (and will likely support more than that too).

This will allow the correct error message to flow through DirectML to WinMLRunner.

### DIFF
--- a/Tools/WinMLRunner/src/Run.cpp
+++ b/Tools/WinMLRunner/src/Run.cpp
@@ -293,13 +293,6 @@ HRESULT CheckIfModelAndConfigurationsAreSupported(LearningModel& model, const st
         else if (inputFeature.Kind() == LearningModelFeatureKind::Tensor)
         {
             auto tensorFeatureDescriptor = inputFeature.try_as<TensorFeatureDescriptor>();
-            if (tensorFeatureDescriptor.Shape().Size() > 4 && deviceType != DeviceType::CPU)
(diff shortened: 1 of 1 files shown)


Tools/WinMLRunner/src/Run.cpp
```diff
@@ -293,13 +293,6 @@ HRESULT CheckIfModelAndConfigurationsAreSupported(LearningModel& model, const st
         else if (inputFeature.Kind() == LearningModelFeatureKind::Tensor)
         {
             auto tensorFeatureDescriptor = inputFeature.try_as<TensorFeatureDescriptor>();
-            if (tensorFeatureDescriptor.Shape().Size() > 4 && deviceType != DeviceType::CPU)
```
smk2007, line 296: should you leave this check and update the check and message to allow up to 5d tensors and not 6?

fdwr: I'm adding 6D, 7D, 8D in just a few weeks. Please don't put artificial limitations in a tool, when the API should check it.

PR #330, opened 2020-07-17
User/t limay/style transfer framerate

linnealovespie:
* Added frame rate measurement as a the time it between the start of successive ProcessFrame calls.
* Fixed the bugs with CPU webstreaming hanging. 
* All the other updates like rewriting CPP component, since that hasn't been merged with my main branch yet.


Samples/StyleTransfer/VideoEffect/StyleTransferEffectCpp/StyleTransferEffect.cpp
```diff
@@ -0,0 +1,94 @@
...
+		VideoFrame outputFrame = context.OutputFrame();
+
+		std::lock_guard<mutex> guard{ Processing };
+		OutputDebugString(L"PF Locked | ");
+		Binding.Bind(InputImageDescription, inputFrame);
+		Binding.Bind(OutputImageDescription, outputTransformed);
+
+		OutputDebugString(L"PF Eval | ");
+		Session.Evaluate(Binding, L"test");
+		outputTransformed.CopyToAsync(outputFrame).get();
```
tiagoshibata, line 64: Is `CopyToAsync` doing any transformations? If not you could put it directly into context.OutputFrame()

Samples/StyleTransfer/VideoEffect/StyleTransferEffectCpp/StyleTransferEffect.cpp
```diff
@@ -0,0 +1,94 @@
...
+	StyleTransferEffect::StyleTransferEffect() : outputTransformed(VideoFrame(Windows::Graphics::Imaging::BitmapPixelFormat::Bgra8, 720, 720)),
+		Session(nullptr),
+		Binding(nullptr)
+	{
+	}
+
+	IVectorView<VideoEncodingProperties> StyleTransferEffect::SupportedEncodingProperties() {
+		VideoEncodingProperties encodingProperties = VideoEncodingProperties();
+		encodingProperties.Subtype(L"ARGB32");
+		return single_threaded_vector(std::move(std::vector<VideoEncodingProperties>{encodingProperties})).GetView();
```
tiagoshibata, line 19: Just FYI, std::move doesn't make a difference here (it's already moving anyways, since you're giving std::vector as a rvalue)

Samples/StyleTransfer/VideoEffect/StyleTransferEffectCpp/StyleTransferEffect.cpp
```diff
@@ -0,0 +1,94 @@
...
+		if (firstProcessFrameCall) {
+			m_StartTime = now;
+			firstProcessFrameCall = false;
+		}
+		// On the second and any proceding process, 
+		else {
+			timePassed = std::chrono::duration_cast<std::chrono::milliseconds>(now - m_StartTime);
+			m_StartTime = now;
+			Notifier.SetFrameRate(1000.f / timePassed.count()); // Convert to FPS: milli to seconds, invert 
+		}
```
tiagoshibata, line 50: These should be inside the lock, or use some kind of thread safe types.

Samples/StyleTransfer/VideoEffect/StyleTransferEffectCpp/StyleTransferEffect.cpp
```diff
@@ -0,0 +1,94 @@
...
+	void StyleTransferEffect::SetProperties(IPropertySet config) {
+		this->configuration = config;
+		hstring modelName;
+		IInspectable val = config.TryLookup(L"ModelName");
+		if (!val) {
+			return;
+		}
+		modelName = unbox_value<hstring>(val);
+		val = configuration.TryLookup(L"UseGPU");
+		bool useGpu = unbox_value<bool>(val);
```
tiagoshibata, line 82: Should it check if val is valid first?

Samples/StyleTransfer/VideoEffect/StyleTransferEffectCpp/StyleTransferEffect.cpp
linnealovespie: I use CopyToAsync for auto-cropping, since the output from Session.Evaluate is a VideoFrame of size 720x720, and context.OutputFrame() is a VideoFrame of size 640x360. I'm taking a second stab at whether I can resize the mediacapture itself to fit the dimensions of the model.

tiagoshibata: Is this the behavior you see? VideoFrame.CopyToAsync doesn't seem to resize:

> The target frame must have the same pixel format, width, height, and alpha mode as the source frame.

(from https://docs.microsoft.com/en-us/uwp/api/windows.media.videoframe.copytoasync?view=winrt-19041 . Could be that the docs are outdated though)

PR #337, opened 2020-08-06
Change winmlrunner to accept LearningModelSessionOptions through run function

ryanlai2:
This will allow apps that consume WinMLRunner static library to be able to modify their own LearningModelSessionOptions.

### DIFF
--- a/Tools/WinMLRunner/src/main.cpp
+++ b/Tools/WinMLRunner/src/main.cpp
@@ -5,6 +5,20 @@
 #include <codecvt>
 using namespace std;
 
+void PopulateSessionOptions(LearningModelSessionOptions& sessionOptions)
+{
+    // Batch Size Override as 1
+    try
+    {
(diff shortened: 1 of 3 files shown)


Tools/WinMLRunner/src/main.cpp
```diff
@@ -36,7 +50,9 @@ int main(int argc, char *argv[])
         wprintf(error.message().c_str());
         return error.code();
     }
-    int returnCode = run(*commandLineArgs, profiler, deviceList);
+    LearningModelSessionOptions sessionOptions;
+    PopulateSessionOptions(sessionOptions);
+    int returnCode = run(*commandLineArgs, profiler, deviceList, sessionOptions);
     free(commandLineArgs);
```
tiagoshibata, line 56: Not related to your PR, but just saw this big red flag :( I can fix this in another PR

PR #342, opened 2020-08-21
Check in sample showcasing Rust projection of Winrt 

ryanlai2:
This sample showcases WinML using Squeezenet through the Rust WinRT projection. It utilizes Microsoft.AI.MachineLearning 1.4.0 Nuget package.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "Tools/WinML-Dashboard/deps/Netron"]
 	path = Tools/WinMLDashboard/deps/Netron
 	url = https://github.com/lutzroeder/Netron.git
-[submodule ".\\Samples\\RustSqueezenet\\winrt-rs"]
-	path = .\\Samples\\RustSqueezenet\\winrt-rs
+[submodule "Samples/RustSqueezenet/winrt-rs"]
+	path = Samples/RustSqueezenet/winrt-rs
 	url = https://github.com/microsoft/winrt-rs.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "Tools/WinML-Dashboard/deps/Netron"]
 	path = Tools/WinMLDashboard/deps/Netron
 	url = https://github.com/lutzroeder/Netron.git
+[submodule ".\\Samples\\RustSqueezenet\\winrt-rs"]
+	path = .\\Samples\\RustSqueezenet\\winrt-rs
+	url = https://github.com/microsoft/winrt-rs.git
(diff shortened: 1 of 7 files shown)


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "Tools/WinML-Dashboard/deps/Netron"]
 	path = Tools/WinMLDashboard/deps/Netron
 	url = https://github.com/lutzroeder/Netron.git
+[submodule "Samples/RustSqueezenet/winrt-rs"]
```
tiagoshibata, line 4: Why build a specific commit instead of using the winrt package?

Samples/RustSqueezenet/copy_collateral.bat
```diff
@@ -0,0 +1,3 @@
+xcopy ..\..\SharedContent\models\SqueezeNet.onnx .\target\debug\ /Y
```
tiagoshibata, line 1: robocopy instead of xcopy prevents unnecessary copies. Would be nice if it worked in release too, and if it was part of build.rs instead of a separate bat.

.gitmodules
ryanlai2: There are features that aren't released in the official package yet that are enabled in the latest version of the github repository. I explained it in the readme.md

Samples/RustSqueezenet/copy_collateral.bat
ryanlai2: Currently there isn't a good way to incorporate copying resource files during cargo build. You can see in issues:
- https://github.com/rust-lang/cargo/issues/5305
- https://github.com/rust-lang/cargo/issues/1759#issuecomment-391694959

ryanlai2: I'm asking winrt-rs folks, just to be sure.

.gitmodules
ryanlai2: I'm asking winrt-rs folks, just to be sure.

ryanlai2: Yeah, without referencing the submodule, bind doesn't work as mentioned in the issue I linked.

ryanlai2: Issue filed: https://github.com/microsoft/winrt-rs/issues/318
It is a legit but acknowledge by winrt-rs team.

zhiburt (commented):
I've run into this issue from you link on rust-lang/cargo#5305.
I would say you did a good job bypassing this :+1: for binaries even though it seems don't work for libraries.

And accidentally just left a subtle review comments :smile:

PS: I know that it meaningless but as I went through code anyway I decided to left it.

PR #348, opened 2020-09-04
Build MicrosoftMLRunner for using WinML nuget and make WinMLRunner only load from System32

ryanlai2:
There are a few changes here:
- Add ifdef USE_WINML_NUGET to switch between using Windows.AI.Machinelearning and Microsoft.AI.MachineLearning
  - Separate out the build configurations to be able to build debug and release versions of inbox and nuget. For example : Release_Inbox and Release_Nuget.
  - For nuget builds, build MicrosoftMLRunner.exe instead of WinMLRunner.exe
- Update WinMLRunner README.md documentation to explain usage of nuget in MicrosoftMLRunner.
- Remove dllload.cpp to remove WinRT_RoGetActivationFactory hook. Don't want to advocate this type of DLL hooking.
- Refactor OutputHelper.h and BindingUtilities.h by moving all of the implementation into respective .cpp files and only exposing functions
- Add WinML nuget 1.4.0 and Winrt nuget
- Fix WinMLRunner tests to use const cast


Tools/WinMLRunner/src/BindingUtilities.cpp
```diff
@@ -0,0 +1,1004 @@
+#include "d3dx12.h"
```
ryanlai2, line 1: Implementation was moved from BindingUtilties.h into BindingUtilities.cpp

Tools/WinMLRunner/src/OutputHelper.cpp
```diff
@@ -0,0 +1,1197 @@
+#include "Common.h"
```
ryanlai2, line 1: Implementation was moved from OutputHelper.h into OutputHelper.cpp

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
```diff
@@ -865,7 +865,7 @@ namespace WinMLRunnerTest
         TEST_METHOD(LoadModelFailModelNotFound)
         {
             const std::wstring command = BuildCommand({ EXE_PATH, L"-model", L"invalid_model_name" });
-            Assert::AreEqual(HRESULT_FROM_WIN32(E_FAIL), RunProc((wchar_t *)command.c_str()));
+            Assert::AreEqual(HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND), RunProc((wchar_t *)command.c_str()));
```
tiagoshibata, line 868: Why is the cast needed?

Tools/WinMLRunner/src/BindingUtilities.cpp
```diff
@@ -0,0 +1,1004 @@
...
+#include <random>
+#include <time.h>
+#include "Windows.AI.Machinelearning.Native.h"
```
tiagoshibata, line 4: Is this header inbox only?

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
ryanlai2: .c_str() produces const wchar_t* and RunProc takes in LPWSTR which is needed for CreateProcessW : https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw

Tools/WinMLRunner/src/BindingUtilities.cpp
ryanlai2: Good point! I fixed it so that the nuget version of native header is used when nuget configuration build is specified.

Testing/WinMLRunnerTest/WinMLRunnerTest.cpp
tiagoshibata: > (wchar_t *)command.c_str()

> .c_str() produces const wchar_t*

Aren't you casting wchar_t* to wchar_t*, then?

tiagoshibata: Oh, const cast, got it... though it's better to use const_cast

ryanlai2: Oh, didn't know about const_cast. Thanks, let me change it.

Tools/WinMLRunner/src/OutputHelper.h
```diff
@@ -62,1162 +22,50 @@ class OutputHelper
...
-            }
-        }
-        else
-        {
-            std::cout << "Failed to Print Learning Model Device Information" << std::endl;
-        }
-    }
-
+                             const std::string& status) const;
+    void PrintModelInfo(std::wstring modelPath, LearningModel model) const;
```
tiagoshibata, line 32: Should take const references

Tools/WinMLRunner/src/OutputHelper.h
```diff
@@ -62,1162 +22,50 @@ class OutputHelper
...
-    std::vector<double> m_clockBindTimes;
-    std::vector<double> m_clockEvalTimes;
-
-    std::wstring getCsvFileNamePerIterationResult() { return m_csvFileNamePerIterationResult; }
-#if defined(_AMD64_)
+                                   const std::vector<std::pair<std::string, std::string>>& perfFileMetadata) const;
+    static void PrintLearningModelDevice(const LearningModelDeviceWithMetadata& device);
+    static std::wstring FeatureDescriptorToString(const ILearningModelFeatureDescriptor& descriptor);
+    static bool doesDescriptorContainFP16(const ILearningModelFeatureDescriptor& descriptor);
+    static bool doesModelContainFP16(const LearningModel model);
```
tiagoshibata, line 57: Should take const reference

tiagoshibata (approved):
Looks good to me, left a couple of comments on the refactored code

PR #354, opened 2020-10-05
fix access violation bug that occurs when csv file is too big

ryanlai2:
WinMLRunner run into access violation error in the scenario where the csv file tensor is larger than what is expected for the model


fdwr (approved):
Thanks.

PR #369, opened 2021-01-02
Include documentation for updating WinMLDashboard + executing these steps

orilevari:
Include documentation for updating WinMLDashboard in UPDATING.md

This PR also follows the steps in this document.

Side notes:
1) protobuf is updated due to a new requirement in tf2onnx that breaks support with the previous version of protobuf.
2) "optional" dependency updates are not included in this PR because these updates break a lot of conversion paths and should be tackled in a future PR.


Tools/WinMLDashboard/public/convert.py
```diff
@@ -87,7 +87,7 @@ def libSVM_converter(args):
                                  target_opset=get_opset(args.ONNXVersion))
     return onnx_model
 
-def convert_tensorflow_file(filename, opset, input_names, output_names):
+def convert_tensorflow_file(filename, opset, output_names):
```
tiagoshibata, line 90: Should line 100 be `converted_model = onnxmltools.convert_tensorflow(..., input_names=input_names, ...)` ?

orilevari: The input names aren't needed for calling into the tensorflow conversion. In this previous PR (https://github.com/microsoft/Windows-Machine-Learning/pull/361/files) I mistakenly added this parameter to the convert_tensorflow_file function, and so I removed it now.

tiagoshibata (approved):

PR #371, opened 2021-01-05
add documentation for updating winmlrunner

orilevari:


Tools/WinMLRunner/UPDATING.md
```diff
@@ -0,0 +1,12 @@
+# Updating WinMLRunner
+
+WinMLRunner should be updated in cadence with updates to WinML and OnnxRuntime.
```
martinb35, line 3: We are trying not to use the word "WinML" and prefer "Windows ML" or "Windows Machine Learning" (and also "ONNX Runtime" with a space in between the words). Also, what about linking to the onnx runtime repo here to make it clear what you're referring to?

martinb35 (approved):

PR #376, opened 2021-02-01
add yolov4 object detection sample

zhangxiang1993:


Samples/YOLOv4ObjectDetection/YOLOv4ObjectDetection/App.xaml.cs
```diff
@@ -0,0 +1,100 @@
...
+            if (rootFrame == null)
+            {
+                // Create a Frame to act as the navigation context and navigate to the first page
+                rootFrame = new Frame();
+
+                rootFrame.NavigationFailed += OnNavigationFailed;
+
+                if (e.PreviousExecutionState == ApplicationExecutionState.Terminated)
+                {
+                    //TODO: Load state from previously suspended application
```
martinb35, line 55: It looks like a lot of these comments were automatically added when the solution was created. Can you look through all the comments and edit or remove the ones that don't apply? For example, this one is a "TODO" - are we actually going to do something here or can we change/remove the comment?

Samples/YOLOv4ObjectDetection/README.md
```diff
@@ -0,0 +1,27 @@
+# YOLOv4 Object Detection Sample
+
+This UWP C# app uses the YOLOv4 model to detect objects in video streams. 
+
+The [YOLOv4 model](https://github.com/hunglc007/tensorflow-yolov4-tflite) is trained with [Tensorflow-DML](https://github.com/microsoft/tensorflow-directml), and converted to ONNX through [tf2onnx](https://github.com/onnx/tensorflow-onnx)
```
martinb35, line 5: do we need to add any licensing text due to including the yolo model?

Samples/YOLOv4ObjectDetection/YOLOv4ObjectDetection/MainPage.xaml.cs
```diff
@@ -0,0 +1,372 @@
...
+            ProcessFrame();
+        }
+
+        private float Sigmoid(float val)
+        {
+            var x = (float)Math.Exp(val);
+            return x / (1.0f + x);
+        }
+
+        private float ComputeIOU(DetectionResult DRa, DetectionResult DRb)
```
martinb35, line 192: what does "IOU" mean? Would it be better to use the full words?

Samples/YOLOv4ObjectDetection/YOLOv4ObjectDetection/MainPage.xaml.cs
```diff
@@ -0,0 +1,372 @@
...
+using Windows.UI.Xaml.Data;
+using Windows.UI.Xaml.Input;
+using Windows.UI.Xaml.Media;
+using Windows.UI.Xaml.Media.Imaging;
+using Windows.UI.Xaml.Navigation;
+using Windows.UI.Xaml.Shapes;
+
+// The Blank Page item template is documented at https://go.microsoft.com/fwlink/?LinkId=402352&clcid=0x409
+
+namespace YOLOv4ObjectDetection
```
martinb35, line 28: I think this is based on the AI skills implementation, right? Can you check with Alex if we should link to their repo and mention that it's based on that implementation?

Samples/YOLOv4ObjectDetection/YOLOv4ObjectDetection/MainPage.xaml.cs
```diff
@@ -0,0 +1,372 @@
...
+{
+
+    public sealed partial class MainPage : Page
```
martinb35, line 31: This whole file could use comments around every method and any where there is complexity to explain what is going on.

Samples/YOLOv4ObjectDetection/YOLOv4ObjectDetection/TinyYOLO.cs
```diff
@@ -0,0 +1,47 @@
+// This file was automatically generated by VS extension Windows Machine Learning Code Generator v3
```
martinb35, line 1: are you actually using this file or did it just come in from MLGen? We should remove it if it's not being used.

Samples/YOLOv4ObjectDetection/YOLOv4ObjectDetection/Yolo.cs
```diff
@@ -0,0 +1,47 @@
+// This file was automatically generated by VS extension Windows Machine Learning Code Generator v3
```
martinb35, line 1: is this file being used?

martinb35 (changes requested):

Samples/YOLOv4ObjectDetection/README.md
zhangxiang1993: I think it should be fine. That Github Repo use MIT license. And our Repo is stated to use the same license.

---
In reply to: [568059295](https://github.com/microsoft/Windows-Machine-Learning/pull/376#discussion_r568059295) [](ancestors = 568059295)

Samples/YOLOv4ObjectDetection/YOLOv4ObjectDetection/MainPage.xaml.cs
zhangxiang1993: This is not depend on any AI kills implemetation as I can see. I just renamed the namespace to align StyleTransferObjectDetection in our Repo. Where else do you think it might be related to AI skills?

---
In reply to: [568061428](https://github.com/microsoft/Windows-Machine-Learning/pull/376#discussion_r568061428) [](ancestors = 568061428)

martinb35 (approved):

PR #377, opened 2021-03-03
Add .NET5 Squeezenet Sample

smk2007:
This sample demonstrates how to use .NET5 applications with the Inbox WinML and Redist WinML.


Samples/SqueezeNetObjectDetection/NET5/Labels.json
```diff
@@ -0,0 +1,1002 @@
+{
+  "0": "tench, Tinca tinca",
```
martinb35, line 2: same as https://github.com/microsoft/Windows-Machine-Learning/blob/master/Samples/SqueezeNetObjectDetection/Desktop/cpp/Labels.txt?

can we share?

Samples/SqueezeNetObjectDetection/NET5/Properties/launchSettings.json
```diff
@@ -0,0 +1,9 @@
+{
+  "profiles": {
+    "SqueezeNetObjectDetectionNET5": {
+      "commandName": "Project",
+      "commandLineArgs": "SqueezeNet.onnx kitten_224.png cpu",
```
martinb35, line 5: do we do this in other samples? I think I like it, so it shows something useful right away. maybe we should add it elsewhere.

martinb35 (approved):

Samples/SqueezeNetObjectDetection/NET5/Properties/launchSettings.json
smk2007: yeah, until now this file was part of the gitignore. We should add the relevant copy commands and launchSettings to each sample

Samples/SqueezeNetObjectDetection/NET5/Labels.json
smk2007: We can, currently it is copied into each project. But squeezenet should be refactored to have all common assets pulled out.

PR #398, opened 2021-07-06
Add controls for melspectrogram settings

janezdu:
image


janezdu:
Wrapped up the changes into #398

PR #401, opened 2021-07-23
Enable Running BatchSupport Sample with NuGet

nums11:
This change enables  running the BatchSupport sample with NuGet to avoid the current issues of the BatchSupport sample with Inbox WinML breaking on prior versions of Windows.

### DIFF
--- a/Samples/BatchSupport/BatchSupport.sln
+++ b/Samples/BatchSupport/BatchSupport.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.0.0
+# Visual Studio 15
+VisualStudioVersion = 15.0.28307.1000
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "BatchSupport", "BatchSupport\BatchSupport.vcxproj", "{A43BD199-EC83-4B0E-9950-A2B8AA51439F}"
 	ProjectSection(ProjectDependencies) = postProject
--- a/Samples/BatchSupport/BatchSupport.sln
+++ b/Samples/BatchSupport/BatchSupport.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.28307.1000
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.31515.178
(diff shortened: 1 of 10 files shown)


Samples/BatchSupport/BatchSupport.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.28307.1000
```
tiagoshibata, line 4: I samples are at VS 2017

Samples/SampleSharedLib/SampleSharedLib/SampleSharedLib.vcxproj.filters
```diff
@@ -29,8 +29,8 @@
     <ClCompile Include="FileHelper.cpp">
       <Filter>Source Files</Filter>
     </ClCompile>
-    <ClCompile Include="dllload.cpp">
-      <Filter>Source Files</Filter>
-    </ClCompile>
```
tiagoshibata, line 34: Was it removed on purpose?

Samples/BatchSupport/BatchSupport.sln
nums11: Do I just need to update my version of Visual studio? I thought I was on the latest for Visual Studio 2019

Samples/SampleSharedLib/SampleSharedLib/SampleSharedLib.vcxproj.filters
ryanlai2: Yes, in fact I would even recommend that the dllload.cpp file should just be removed. This was originally used to be able to load our inbox Windows.AI.Machinelearning.dll from local folder as the app. However, now we have NuGet as a solution, we don't want to have this backdoor way of loading our inbox binaries in our sample.

Samples/BatchSupport/BatchSupport.sln
ryanlai2: I'm fine with making this sample Visual Studio 2019 but then we'll need to update the README documentation of this sample to indicate that 2019 is needed.

nums11: I'm fine with keeping it at VS 2017 to be consistent with the rest of the samples but I'm not sure how to edit this file. From what I understand this is solution file is generated and used by Visual Studio so it's not regularly editable. I figure I could edit in NotePad or a regular text editor but is that the best solution?

ryanlai2: Yeah just edit solution file.

Please keep the original : # Visual Studio 15
VisualStudioVersion = 15.0.28307.1000

tiagoshibata (approved):

PR #402, opened 2021-07-28
revise finding the maximum energy of the label code

BEOKS:
I think that updating 'probability' instance is missed while finding the maximum "energy" of the label


martinb35 (approved):

martinb35:
thanks for fixing this!

PR #411, opened 2021-09-09
Add support for user specified inputNames, bumped tf2onnx version to …

JingqiaoFu:
…support op set 14, fixed op set version. Note that InputName is optional. Shape override currently supported in tf2onnx is not yet supported in onnxmltools, so that's not added to WinMLDashboard yet, optional InputNames will open doors to that change in the future.

### DIFF
--- a/Tools/WinMLDashboard/public/convert.py
+++ b/Tools/WinMLDashboard/public/convert.py
@@ -30,6 +30,8 @@ def get_opset(ONNXVersion):
         return 7
     elif '1.3' == ONNXVersion:
         return 8
+    elif '1.4' == ONNXVersion:
+        return 9
     elif '1.5' == ONNXVersion:
         return 10
     elif '1.6' == ONNXVersion:
--- a/Tools/WinMLDashboard/public/convert.py
+++ b/Tools/WinMLDashboard/public/convert.py
@@ -41,6 +41,7 @@ def get_opset(ONNXVersion):
     elif '1.9' == ONNXVersion:
         return 14
     else:
+        print('WARNING: ONNX Version ' + ONNXVersion + ' does not map to any known opset version, defaulting to opset V7')
         return 7
 
 
--- a/Tools/WinMLDashboard/public/convert.py
+++ b/Tools/WinMLDashboard/public/convert.py
@@ -7,7 +7,7 @@ def parse_args():
     parser.add_argument('source', help='source  model')
     parser.add_argument('framework', help='source framework model comes from')
     parser.add_argument('ONNXVersion', help='which ONNX Version to convert to')
-    parser.add_argument('inputNames', help='names of input nodes')
+    parser.add_argument('inputNames', help='names of input nodes (optional)')
     parser.add_argument('outputNames', help='names of output nodes')
     parser.add_argument('destination', help='destination ONNX model (ONNX or prototxt extension)')
     parser.add_argument('--name', default='WimMLDashboardConvertedModel', help='(ONNX output only) model name')
--- a/Tools/WinMLDashboard/public/convert.py
+++ b/Tools/WinMLDashboard/public/convert.py
@@ -7,7 +7,7 @@ def parse_args():
     parser.add_argument('source', help='source  model')
     parser.add_argument('framework', help='source framework model comes from')
     parser.add_argument('ONNXVersion', help='which ONNX Version to convert to')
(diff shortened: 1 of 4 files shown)


Tools/WinMLDashboard/public/convert.py
```diff
@@ -7,6 +7,7 @@ def parse_args():
     parser.add_argument('source', help='source  model')
     parser.add_argument('framework', help='source framework model comes from')
     parser.add_argument('ONNXVersion', help='which ONNX Version to convert to')
+    parser.add_argument('inputNames', help='names of input nodes (optional)')
```
tiagoshibata, line 10: The argument isn't optional as-is, maybe make it optional or change the doc to `names of input nodes (or empty string)`

Tools/WinMLDashboard/public/convert.py
```diff
@@ -87,22 +96,24 @@ def libSVM_converter(args):
...
 
-def convert_tensorflow_file(filename, opset, output_names):
+def convert_tensorflow_file(filename, opset, input_names, output_names):
     import tensorflow
     from tensorflow.core.framework import graph_pb2
     from tensorflow.python.tools import freeze_graph
     import onnx
     import tensorflow as tf
 
+    print(input_names)
```
tiagoshibata, line 106: You forgot a print here

Tools/WinMLDashboard/public/convert.py
```diff
@@ -31,6 +32,14 @@ def get_opset(ONNXVersion):
...
+    elif '1.6' == ONNXVersion:
+        return 11
+    elif '1.7' == ONNXVersion:
+        return 12
+    elif '1.8' == ONNXVersion:
+        return 13
+    elif '1.9' == ONNXVersion:
+        return 14
     else:
         return 7
```
tiagoshibata, line 44: Maybe warn about the fallback:

```py
def get_opset(onnx_version):
    version_to_opset = {
        '1.2': 7,
        '1.3': 8,
        '1.5': 10,
        '1.6': 11,
        '1.7': 12,
        '1.8': 13,
        '1.9': 14,
    }
    opset = version_to_opset.get(onnx_version)
    if not opset:
        # print some warning
        return 7
    return opset
```

(Also, is there no opset 9?)

tiagoshibata (changes requested):

Tools/WinMLDashboard/public/convert.py
JingqiaoFu: Oh yeah, thanks

Tools/WinMLDashboard/public/convert.py
JingqiaoFu: Will add required=False

Tools/WinMLDashboard/public/convert.py
JingqiaoFu: I believe there should be? I don't have the context of why it's skipped here so I didn't try to add it, maybe someone in the team can fill me in here?

Tools/WinMLDashboard/public/convert.py
JingqiaoFu: it looks like all the options are optional by default, so this should be OK as is?

tiagoshibata: At least destination (the last parameter) should be non-empty I think. You'd need a non-positional argument (like `--inputName`) or having inputNames in the end of the argument list

JingqiaoFu: hmmm, I have tested it and it was working as is, are you commenting because you believe this won't work as expected or because it's not a good practice?

Tools/WinMLDashboard/public/convert.py
JingqiaoFu: Added v9 support

Tools/WinMLDashboard/public/convert.py
tiagoshibata: Where is the model saved to if the destination is empty? I though it was a required parameter

tiagoshibata (approved):

Tools/WinMLDashboard/public/convert.py
JingqiaoFu: When the conversion is done, a file dialog will be opened to let the user choose destination & file name

PR #439, opened 2021-11-03
Extend Github Public Samples Build Timeout to 90 minutes

nums11:
This change extends the GitHub Public Samples Build Timeout to 90 minutes to counteract the increased execution time due to the OpenCV Build Task
- All the tasks were placed underneath a job so that the timeoutInMinutes flag could be added globally

### DIFF
--- a/azure-pipelines-samples.yml
+++ b/azure-pipelines-samples.yml
@@ -1,34 +1,17 @@
 resources:
-
 - repo: self
   clean: true
-
-variables:
-  SamplesBin: SamplesBin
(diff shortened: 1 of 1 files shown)


martinb35:
why does it look like the whole file changed? is there a whitespace or line-ending change?

nums11:
> why does it look like the whole file changed? is there a whitespace or line-ending change?

All the tasks were placed underneath a job so that the timeoutInMinutes flag could be added globally

martinb35 (approved):

PR #440, opened 2021-11-03
Use batched SqueezeNet model in batching sample

nums11:
This change adds the batched SqueezeNet model to the models folder in the WinMLSamplesGallery and updates the batching sample to use that model.
- This also fixes the issues with the batching sample crashing

### DIFF
--- a/Samples/WinMLSamplesGallery/WinMLSamplesGallery/WinMLSamplesGallery.csproj
+++ b/Samples/WinMLSamplesGallery/WinMLSamplesGallery/WinMLSamplesGallery.csproj
@@ -153,6 +153,7 @@
   </ItemGroup>
 
   <ItemGroup>
+	<Folder Include="Models\" />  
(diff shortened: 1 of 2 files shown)


Samples/WinMLSamplesGallery/WinMLSamplesGallery/WinMLSamplesGallery.csproj
```diff
@@ -151,7 +153,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Folder Include="Models\" />
```
smk2007, line 154: Can we keep this? Does it need to go?

nums11: Yea not sure why this was removed.

smk2007 (approved):

PR #443, opened 2021-11-11
MFT stub

linnealovespie:
1. Create a stub MFT in TransformBlur.cpp
2. Implement image invert in SegmentModel.cpp using LearningModelBuilder as a stand-in for image segmentation
3. Replace default Enhanced Video Renderer (EVR) with DX11VideoRenderer in order to use D3D11 resources to render to the output window
4. Handle extracting the DXGIDeviceManager from the DX11VideoRender and sharing it with the MFT so that it can be D3D-Aware

The bulk of the logic happens in the following functions: 
1. Player.cpp::AddBranchToPartialTopology
2. TransformBlur::ProcessOutput
3. SegmentModel::RunTest


Samples/BackgroundBlur/BackgroundBlur/SegmentModel.h
```diff
@@ -0,0 +1,40 @@
...
+private: 
+	LearningModel GetBackground(long n, long c, long h, long w);
+	LearningModel GetForeground(long n, long c, long h, long w);
+	LearningModel Argmax(long axis, long h, long w);
+	LearningModel FCNResnet();
+	LearningModel ReshapeFlatBufferToNCHW(long n, long c, long h, long w);
+	LearningModel ReshapeFlatBufferToNCHWAndInvert(long n, long c, long h, long w);
+	LearningModel Normalize0_1ThenZScore(long height, long width, long channels, std::vector<float> means, std::vector<float> stddev);
+
+	LearningModelBinding Evaluate(LearningModelSession sess, std::vector<ITensor*> input, ITensor* output, bool wait = false);
```
JingqiaoFu, line 33: I think we should avoid passing the actual object here, best to use reference or smart pointers?

fdwr: Yeah, C++ passes by value copy. So tis lighter perf and and more flexible to pass an `std::span` rather than a vector copy.

Samples/BackgroundBlur/DX11VideoRenderer/Activate.cpp
```diff
@@ -0,0 +1,214 @@
+#include "pch.h"
```
JingqiaoFu, line 1: So it seems we are copying the DX11VideoRenderer, possibly from here: https://github.com/microsoft/Windows-classic-samples/tree/main/Samples/DX11VideoRenderer ?
Do you think we could maybe build it into a lib and include .lib and .h files? or submoduling this directory could also be an option, I'm just wondering if we really need to expose(and maintain a separate copy) the source code for this renderer in our sample

Samples/BackgroundBlur/.gitignore
```diff
@@ -0,0 +1 @@
+*.onnx
```
martinb35, line 1: where is the model coming from? this seems to suggest you aren't checking it in.

Samples/BackgroundBlur/BackgroundBlur/BackgroundBlur.cpp
```diff
@@ -0,0 +1,511 @@
...
+    default:
+        return DefWindowProc(hwnd, message, wParam, lParam);
+    }
+    return 0;
+}
+
+//  Open an audio/video file.
+void OnFileOpen(HWND hwnd)
+{
+    IFileOpenDialog* pFileOpen = NULL;
```
martinb35, line 178: should these be wrapped in smart pointers so you don't have to manually release the memory?

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
```diff
@@ -0,0 +1,402 @@
...
+	ONNX_UINT64 = 13,
+	ONNX_COMPLEX64 = 14,     // complex with float32 real and imaginary components
+	ONNX_COMPLEX128 = 15,    // complex with float64 real and imaginary components
+
+	// Non-IEEE floating-point format based on IEEE754 single-precision
+	// floating-point number truncated to 16 bits.
+	// This format has 1 sign bit, 8 exponent bits, and 7 mantissa bits.
+	ONNX_BFLOAT16 = 16,
+}OnnxDataType;
+hstring modelPath = L"C:\\Windows-Machine-Learning\\Samples\\BackgroundBlur\\BackgroundBlur\\Assets\\fcn-resnet50-11.onnx";
```
martinb35, line 36: the hardcoded path probably won't work on other people's machines.

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
```diff
@@ -0,0 +1,402 @@
...
+bool useGPU = true;
```
martinb35, line 37: do these need to be global variables?

Samples/BackgroundBlur/.gitignore
linnealovespie: I'm not checking it in because it's too big to commit. I'll add a download task build events so that it downloads the onnx file if it doesn't already exist in the assets.

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.h
JingqiaoFu: Wasn't aware that WinML objs are com objs and therefore they are access with com ptr

Samples/BackgroundBlur/BackgroundBlur/BackgroundBlur.cpp
linnealovespie: This is all from demo code, but I can refactor to use smart pointers instead if it's the better coding practice to demonstrate.

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
linnealovespie: Set to be a member variable of SegmentModel

Samples/BackgroundBlur/DX11VideoRenderer/Activate.cpp
linnealovespie: I've had to make some changes to the DX11VideoRenderer (you're correct that it's from the mentioned repo), so I don't think we can submodule the directory and immediately use it in the project.  The big one is that the sample uses C++14 and I've had to make some changes (eg. using typename modifiers to in templated classes) to get it compatible with C++17.

Samples/BackgroundBlur/BackgroundBlur/BackgroundBlur.cpp
```diff
@@ -0,0 +1,515 @@
+// THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF
+// ANY KIND, EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO
+// THE IMPLIED WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A
+// PARTICULAR PURPOSE.
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
tiagoshibata, line 6: I believe all samples are under the MIT license

Samples/BackgroundBlur/BackgroundBlur/BackgroundBlur.cpp
```diff
@@ -0,0 +1,515 @@
...
+
+    // Register the window class.
+    ZeroMemory(&wcex, sizeof(WNDCLASSEX));
+    wcex.cbSize = sizeof(WNDCLASSEX);
+    wcex.style = CS_HREDRAW | CS_VREDRAW;
+    wcex.lpfnWndProc = WndProc;
+    wcex.hInstance = hInst;
+    wcex.hbrBackground = (HBRUSH)(COLOR_WINDOW + 1);
+    wcex.lpszMenuName = MAKEINTRESOURCE(IDC_MFPLAYBACK);
+    wcex.lpszClassName = szWindowClass;
```
tiagoshibata, line 99: Nipick: In C++, unspecified members of a struct are initialized to zero, so:

```cpp
WNDCLASSEX wcex {};  // All members are initialized to zero, no need for ZeroMemory
```

Or even better:

```cpp
    WNDCLASSEX wcex {
        .cbSize = sizeof(WNDCLASSEX);
        .style = CS_HREDRAW | CS_VREDRAW;
        .lpfnWndProc = WndProc;
        .hInstance = hInst;
        .hbrBackground = (HBRUSH)(COLOR_WINDOW + 1);
        .lpszMenuName = MAKEINTRESOURCE(IDC_MFPLAYBACK);
        .lpszClassName = szWindowClass;
    };
    // These members are initialized, any remaining members are zero
```

Samples/BackgroundBlur/BackgroundBlur/BackgroundBlur.cpp
```diff
@@ -0,0 +1,515 @@
...
+    if (SUCCEEDED(hr))
+    {
+        UpdateUI(hwnd, Closed);
+        return 0;   // Success.
+    }
+    else
+    {
+        NotifyError(NULL, L"Could not initialize the player object.", hr);
+        return -1;  // Destroy the window
+    }
```
tiagoshibata, line 297: Nitpick:

```cpp
    if (!SUCCEEDED(hr))
    {
        NotifyError(NULL, L"Could not initialize the player object.", hr);
        return -1;  // Destroy the window
    }
    UpdateUI(hwnd, Closed);
    return 0;   // Success.
```

Samples/BackgroundBlur/BackgroundBlur/BackgroundBlur.cpp
```diff
@@ -0,0 +1,515 @@
...
+        break;
+
+    case Started:
+        bPlayback = TRUE;
+        break;
+
+    case Paused:
+        bPlayback = TRUE;
+        break;
+    }
```
tiagoshibata, line 384: There's a duplicate case:

```cpp
    switch (state)
    {
    case OpenPending:
        bWaiting = TRUE;
        break;

    case Started:
    case Paused:
        bPlayback = TRUE;
        break;
    }
```

Samples/BackgroundBlur/BackgroundBlur/BackgroundBlur.cpp
```diff
@@ -0,0 +1,515 @@
...
+        return (INT_PTR)TRUE;
+
+    case WM_COMMAND:
+        switch (LOWORD(wParam))
+        {
+        case IDOK:
+            if (pUrl)
+            {
+                // Get the URL from the edit box in the dialog. This function 
+                // allocates memory. The caller must call CoTaskMemAlloc.
```
tiagoshibata, line 438: Comment looks misleading (caller must call CoTaskMemFree, right?). Also it seems to be leaking the allocation, I think you'd have to check if pUrl->pszURL is not null and free it before calling AllocGetWindowText.

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
```diff
@@ -0,0 +1,393 @@
...
+	{
+		// v1: just get the reference- should fail
+		auto reference = foreground.as<TensorUInt8Bit>().CreateReference().data();
+
+		// v3: get from a d3dresource
+		/*ID3D12Resource* res = NULL;
+		HRESULT hr = foreground.as<ITensorNative>()->GetD3D12Resource(&res);
+		UINT DstRowPitch = 0, DstDepthPitch = 0, SrcSubresource = 0;
+		hr = res->ReadFromSubresource((void*)*pDest, DstRowPitch, DstDepthPitch, SrcSubresource, NULL);*/
+		return;
```
tiagoshibata, line 106: Is this to be done?

Samples/BackgroundBlur/BackgroundBlur/BackgroundBlur.cpp
linnealovespie: I can't use designated initializers (eg. the second option) with C++17, so I'll go with the first option.

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
```diff
@@ -0,0 +1,393 @@
...
+}
+
+LearningModel SegmentModel::Argmax(long axis, long h, long w)
+{
+	auto builder = LearningModelBuilder::Create(12)
+		.Inputs().Add(LearningModelBuilder::CreateTensorFeatureDescriptor(L"Data", TensorKind::Float, array_view<int64_t const>{ -1, -1, h, w })) // Different input type? 
+		.Outputs().Add(LearningModelBuilder::CreateTensorFeatureDescriptor(L"Output", TensorKind::Float, array_view<int64_t const>{ -1, -1, h, w })) // Output of int64? 
+		.Operators().Add( LearningModelOperator(L"ArgMax")
+			.SetInput(L"data", L"Data")
+			.SetAttribute(L"keepdims", TensorInt64Bit::CreateFromArray(std::vector<int64_t>{1}, std::vector<int64_t>{ 1 }))
```
tiagoshibata, line 154: Nitpick: There's a bunch of places with temporary std::vector. array_view accepts std::array or an initializer_list too, which are much easier and faster to build than a temporary vector that gets thrown away as soon as the call returns.

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
```diff
@@ -0,0 +1,393 @@
...
+			.SetAttribute(L"to", TensorInt64Bit::CreateFromIterable(std::vector<int64_t>{}, std::vector<int64_t>{OnnxDataType::ONNX_FLOAT}))
+			.SetOutput(L"output", L"Output"))
+		;
+
+	return builder.CreateModel();
+}
+
+LearningModel SegmentModel::FCNResnet()
+{
+	auto rel = std::filesystem::current_path();
```
tiagoshibata, line 168: std::filesystem is C++20

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.h
```diff
@@ -0,0 +1,43 @@
+#pragma once
+#include <winrt/Microsoft.AI.MachineLearning.Experimental.h>
+#include <winrt/Microsoft.AI.MachineLearning.h>
+#include <Windows.AI.MachineLearning.native.h>
+#include <strsafe.h>
+#include <wtypes.h>
+#include <winrt/base.h>
+
+using namespace winrt::Microsoft::AI::MachineLearning;
+using namespace winrt::Microsoft::AI::MachineLearning::Experimental;
```
tiagoshibata, line 10: Avoid `using namespace` in headers, it pollutes the namespace of any file that includes this header

tiagoshibata: (As a best practice avoid it in general, but in `cpp` files we tend to use it elsewhere)

Samples/BackgroundBlur/BackgroundBlur/BackgroundBlur.cpp
linnealovespie: This is leftover from the sample I built out from. I can delete and have the code under our MIT license.

Samples/BackgroundBlur/BackgroundBlur/TransformBlur.cpp
```diff
@@ -0,0 +1,1550 @@
...
+{
+    ULONG uCount = InterlockedDecrement(&m_nRefCount);
+    if (uCount == 0)
+    {
+        delete this;
+    }
+    // For thread safety, return a temporary variable.
+    return uCount;
+}
+
```
tiagoshibata, line 109: Isn't there a default implementation of these if you inherit from IUnknown? Just trying to understand why you need to refcount manually.

Also, do we need e.g. copy constructors to update the refcount? It feels like this implementation has gaps

tiagoshibata: (Or make it not copyable)

Samples/BackgroundBlur/BackgroundBlur/TransformBlur.cpp
```diff
@@ -0,0 +1,1550 @@
...
+HRESULT TransformBlur::GetInputAvailableType(
+    DWORD           dwInputStreamID,
+    DWORD           dwTypeIndex, // 0-based
+    IMFMediaType** ppType
+)
+{
+    TRACE((L"GetInputAvailableType (stream = %d, type index = %d)\n", dwInputStreamID, dwTypeIndex));
+
+    AutoLock lock(m_critSec);
```
tiagoshibata, line 371: Which access is meant to be locked? It's better to restrict the lock and make clear what's shared here

Samples/BackgroundBlur/BackgroundBlur/TransformBlur.cpp
```diff
@@ -0,0 +1,1550 @@
...
+
+    // Invoke the image transform function.
+    if (true)
+    {
+        TransformImage_Stub(pDest, pSrc, m_cbImageSize, m_segmentModel);
+    }
+    else
+    {
+        CHECK_HR(hr = E_UNEXPECTED);
+    }
```
tiagoshibata, line 1410: if (true)

Samples/BackgroundBlur/BackgroundBlur/BackgroundBlur.cpp
linnealovespie: is this what you mean? 

`if (pUrl->pszURL) 
                {
                    std::free(pUrl->pszURL);
                }`

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
linnealovespie: yes, in the future this will be allocating d3d12 resources that the video renderer can use directly.

Samples/BackgroundBlur/BackgroundBlur/common/GrowArray.h
```diff
@@ -0,0 +1,122 @@
...
+//-----------------------------------------------------------------------------
+
+
+#pragma once
+
+
+namespace MediaFoundationSamples
+{
+
+    // Class template: Re-sizable array.
```
tiagoshibata, line 20: If there's a reason not to use std::vector, add it to the comments here

Samples/BackgroundBlur/BackgroundBlur/common/PropVar.h
```diff
@@ -0,0 +1,95 @@
...
+// Wrapper class for PROPVARIANT types. 
+//
+// Notes: 
+//    This class offers accessor functions for a subset of the PROPVARIANT
+//    data types, designed to handle all of the types in actual use by 
+//    Media Foundation at this time.
+// 
+///////////////////////////////////////////////////////////////////////////////
+
+class PropVariant : public PROPVARIANT
```
tiagoshibata, line 19: There's std::variant, not sure if it covers your use case

Samples/BackgroundBlur/BackgroundBlur/common/critsec.h
```diff
@@ -0,0 +1,64 @@
+#pragma once
+
```
tiagoshibata, line 2: [It's always better to include what you use](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md)

Samples/BackgroundBlur/BackgroundBlur/common/critsec.h
```diff
@@ -0,0 +1,64 @@
...
+    //////////////////////////////////////////////////////////////////////////
+
+    class CritSec
+    {
+    private:
+        CRITICAL_SECTION m_criticalSection;
+    public:
+        CritSec()
+        {
+            InitializeCriticalSection(&m_criticalSection);
```
tiagoshibata, line 19: Using InitializeCriticalSection and the interlocked methods from Windows is fine, but in general I'd use the methods from the STL (mutex/unique_lock and atomic in this case). They come with C++ RAII, are cross-platform and will call the same or similar Windows APIs under the hood. But no need to refactor all the code now.

Samples/BackgroundBlur/BackgroundBlur/common/mfutils.h
```diff
@@ -0,0 +1,188 @@
...
+        GetStreamCurrentMediaType
+        GetTopoNodeObject
+    */
+
+
+    // MFTimeToMsec: Convert 100-ns time to seconds.
+    inline LONG MFTimeToMsec(const LONGLONG& time)
+    {
+        return (LONG)(time / (ONE_SECOND / ONE_MSEC));
+    }
```
tiagoshibata, line 40: Comments are misleading (1000ns is 1us instead of 1ms, I think all instances of ns should be us)

Samples/BackgroundBlur/BackgroundBlur/common/mfutils.h
```diff
@@ -0,0 +1,188 @@
...
+//
+//////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#include <mftransform.h>
+
+namespace MediaFoundationSamples
+{
+    const MFTIME ONE_SECOND = 10000000; // One second in hns
```
tiagoshibata, line 20: Is this constant correct?

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
linnealovespie: So the parameter type here is winrt::param::iterable, so if I'm reading [Passing parameters into the ABI boundary](https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/pass-parms-to-abi#iterable-parameters) I can pass std::vector and not std::array. Are there any other alternatives that are more effective? If it helps, I can move creating each of these learningmodelsessions out of the RunTest method like I was planning to do once I consolidate the models for pre- and post-processing into a single mega model.

Samples/BackgroundBlur/BackgroundBlur/common/mfutils.h
```diff
@@ -0,0 +1,188 @@
...
+    // This function allocates a buffer and returns it in pszText. The caller must
+    // call CoTaskMemFree on the buffer.
+    //
+    // hwnd:     Handle to the window
+    // pszText:  Receives a pointer to the string.
+    // pcchLen:  Receives the length of the string, in characters, not including
+    //           the terminating NULL character. This parameter can be NULL.
+    //--------------------------------------------------------------------------------------
+
+    inline HRESULT AllocGetWindowText(HWND hwnd, WCHAR **pszText, DWORD *pcchLen)
```
tiagoshibata, line 141: This was duplicated in another file, the other file can probably call it from this header

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
linnealovespie: Looks like it's c++17 as well: https://en.cppreference.com/w/cpp/filesystem/current_path

Samples/BackgroundBlur/BackgroundBlur/BackgroundBlur.cpp
linnealovespie: Removed duplicate case

Samples/BackgroundBlur/DX11VideoRenderer/Activate.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+        hr = pActivate->QueryInterface(IID_PPV_ARGS(ppActivate));
+        if (FAILED(hr))
+        {
+            break;
+        }
+
+        pActivate->m_hwnd = hwnd;
+    }
+    while (FALSE);
```
tiagoshibata, line 37: Is while (FALSE) intended?

Samples/BackgroundBlur/DX11VideoRenderer/Common.h
```diff
@@ -0,0 +1,384 @@
...
+        return t / 10000.0;
+    }
+
+    template <class T> inline T MillisecondsToTicks(const T& t)
+    {
+        return t * 10000;
+    }
+
+    // returns the greatest common divisor of A and B
+    inline int gcd(int A, int B)
```
tiagoshibata, line 75: There's a std::gcd template

tiagoshibata: Oh this was copied from another repo... Feel free to leave it as-is if you don't wanna diverge from the original repo. Less conflicts will make it easier in the future if you want to pull new changes.

Samples/BackgroundBlur/BackgroundBlur/BackgroundBlur.cpp
tiagoshibata: `CoTaskMemFree(pUrl->pszURL);` in the caller (the OnOpenURL function)

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
tiagoshibata: I was looking at the second parameter, which is array_view and can take std::array or std::initializer_list. But from the docs IIterable can also take std::initializer_list, no?

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
tiagoshibata: Oh I was sure it didn't make it into C++17. You're correct, my bad.

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.h
linnealovespie: Makes sense, I can at least move it to SegmentModel.cpp

Samples/BackgroundBlur/DX11VideoRenderer/Common.h
linnealovespie: Yes all of the files in common come from [mediafoundation samples](https://github.com/microsoft/Windows-classic-samples/tree/main/Samples/Win7Samples/multimedia/mediafoundation). I can at least not include the files I'm not using and maybe refactor their sample code later. All of the mf samples are from 5 years ago...

Samples/BackgroundBlur/BackgroundBlur/common/GrowArray.h
linnealovespie: There isn't- in fact the demo doesn't use growarray at all. I'll remove it.

Samples/BackgroundBlur/BackgroundBlur/common/PropVar.h
linnealovespie: Unused code from MF samples, I'll remove from the project.

Samples/BackgroundBlur/BackgroundBlur/common/mfutils.h
linnealovespie: Removed mfutils.h from project because it's not used anywhere.

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
linnealovespie: Oh I see, yes you're right.

PR #448, opened 2021-12-02
Add old samples back to the GitHub Public Samples Build

nums11:
This PR adds the below samples back to the GitHub Public Samples Build and includes fixes for their previous failures
 - MNIST Tutorial CS
 - MNIST UWP CS
 - FNSCandy UWP CS
 - MNIST UWP CPP

Notable fixes include
- NuGet package upgrade from 1.5.2 to latest for the MNIST UWP CS Sample
- Changing the release configuration name in the MNIST UPW CS Sample from 'Nuget' to 'NuGet' to be consistent with the Debug release configuration name

### DIFF
--- a/azure-pipelines-samples.yml
+++ b/azure-pipelines-samples.yml
@@ -134,65 +134,65 @@ jobs:
         createLogFile: true
       condition: and(succeededOrFailed(), eq(variables['BuildConfiguration'], 'Release'))
 
-    # - task: VSBuild@1
-    #   displayName: 'Build UI test solution Testing/**/SamplesTest.sln'
-    #   inputs:
-    #     solution: 'Testing/**/SamplesTest.sln'
-    #     vsVersion: "16.0"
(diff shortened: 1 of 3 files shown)


azure-pipelines-samples.yml
```diff
@@ -134,7 +134,65 @@ jobs:
         createLogFile: true
       condition: and(succeededOrFailed(), eq(variables['BuildConfiguration'], 'Release'))
 
-    # TODO: Add previously failing build tasks
+    - task: VSBuild@1
+      displayName: 'Build UI test solution Testing/**/SamplesTest.sln'
+      inputs:
+        solution: 'Testing/**/SamplesTest.sln'
```
martinb35, line 140: IIRC, this test was supposed to actually run MNIST and draw into the frame - does that still happen somewhere? It looks like this task builds the sample, but does it actually run the samplesTest app somewhere?

nums11: I think it just builds. This was the same task that was there previously; the only change I made was the addition of ' /t:Restore,Clean,Build' to the end of the msbuildArgs.

Maybe there was a task to actually run it in a different pipeline? You can see what the task looked like before I removed it [here](https://github.com/microsoft/Windows-Machine-Learning/pull/422/files#diff-2d5f7450f4896cae0a954e6d1e85749d2815aa90b4edb7fb9ddf110aadf65d54L71).

martinb35 (approved):

PR #467, opened 2022-03-11
Remove Windows-Machine-Learning SampleSharedLib's dependency on BatchSupport sample project

JingqiaoFu:
Remove Windows-Machine-Learning SampleSharedLib's dependency on BatchSupport sample project

### DIFF
--- a/Samples/BatchSupport/BatchSupport/BatchSupport.vcxproj
+++ b/Samples/BatchSupport/BatchSupport/BatchSupport.vcxproj
@@ -44,7 +44,7 @@
   </ItemGroup>
   <PropertyGroup Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>v140</PlatformToolset>
     <PlatformToolset Condition="'$(VisualStudioVersion)' == '15.0'">v141</PlatformToolset>
     <PlatformToolset Condition="'$(VisualStudioVersion)' == '16.0'">v142</PlatformToolset>
     <CharacterSet>Unicode</CharacterSet>
--- a/Samples/BatchSupport/BatchSupport/BatchSupport.vcxproj
+++ b/Samples/BatchSupport/BatchSupport/BatchSupport.vcxproj
@@ -44,7 +44,7 @@
   </ItemGroup>
   <PropertyGroup Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
-    <PlatformToolset>v140</PlatformToolset>
+    <PlatformToolset>v143</PlatformToolset>
     <PlatformToolset Condition="'$(VisualStudioVersion)' == '15.0'">v141</PlatformToolset>
     <PlatformToolset Condition="'$(VisualStudioVersion)' == '16.0'">v142</PlatformToolset>
     <CharacterSet>Unicode</CharacterSet>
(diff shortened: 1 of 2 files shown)


ryanlai2 (dismissed):

linnealovespie:
I think there's some sort of constant circular dependency going on with the samplesSharedLib. Updating SampleSharedLib.vcxproj does successfully remove the dependency, but when I go to build the BatchSupport solution, I get a build error saying that the SamplesSharedLib project can't build with the error: "This project references NuGet package(s) that are missing on this computer.... The missing file is **..\packages\Microsoft.Windows.CppWinRT.2.0.220131.2\build\native\Microsoft.Windows.CppWinRT.props**", even though it's clearly in SamplesSharedLib.vcxproj as seen in this change. 

The only way I've found to get BatchSupport to build is to manually delete these references in SamplesSharedLib.vcxproj, then install the cppwinrt nuget package into BatchSupport the solution itself, thus creating SamplesSharedLib's original dependency on BatchSupport. 

There may be a better way that I don't know about to fix this build issue without re-introducing SamplesSharedLib's dependency on BatchSupport. But if there isn't I suppose we make SamplesSharedLib into just a collection of include-able headers instead of a separate project.

linnealovespie:
I can build BatchSupport now, but I haven't figured out yet why I get the following error when trying to build CustomTensorzation (the other sample that uses SamplesSharedLib): 
LNK2038	mismatch detected for 'C++/WinRT version': value '2.0.201201.7' doesn't match value '2.0.220131.2' in main.obj ..\Samples\CustomTensorization\CustomTensorization\SampleSharedLib.lib(FileHelper.obj)

JingqiaoFu:
> I can build BatchSupport now, but I haven't figured out yet why I get the following error when trying to build CustomTensorzation (the other sample that uses SamplesSharedLib): LNK2038 mismatch detected for 'C++/WinRT version': value '2.0.201201.7' doesn't match value '2.0.220131.2' in main.obj ..\Samples\CustomTensorization\CustomTensorization\SampleSharedLib.lib(FileHelper.obj)

That's really odd, I searched the entire repo for 201201 but found nothing, did you try cleanup build?

linnealovespie:
> That's really odd, I searched the entire repo for 201201 but found nothing, did you try cleanup build?

I did, but it doesn't seem to help. Do you get this error? It could just be something with my local build.

Samples/BatchSupport/BatchSupport/BatchSupport.vcxproj
```diff
@@ -1,199 +1,199 @@
...
-      <Platform>x64</Platform>
-    </ProjectConfiguration>
-    <ProjectConfiguration Include="Release|x64">
-      <Configuration>Release</Configuration>
-      <Platform>x64</Platform>
-    </ProjectConfiguration>
-  </ItemGroup>
-  <PropertyGroup Label="Configuration">
-    <ConfigurationType>Application</ConfigurationType>
-    <PlatformToolset>v140</PlatformToolset>
```
linnealovespie, line 47: OK the linker error I was having seems to be related to not updating platform toolsets to the same version across projects, but it builds once I update them correctly. Is there a reason why BatchSupport is set to the VS2015 toolset? Because all the other projects seem to be VS2019/v141 or above.

JingqiaoFu: No reason for sticking with v140, if that solves your issue, I will upgrade it

JingqiaoFu: Can you list all the projects that you had to upgrade from v140? Want to make sure nothing was missed.

linnealovespie (approved):

PR #469, opened 2022-03-14
Real-time async ML effect Media Foundation Transform

linnealovespie:
1. Implements a CaptureManager to manage the capture engine (ie. the camera) and the capture preview sink (ie. the preview window), maintain a d3d device to use for rendering, and respond to events that come from both the capture engine and user input. --> ```BackgroundBlur/Capture.h```
2. Implements an asynchronous Media Foundation Transform (MFT) called TransformAsync. TransformAsync asynchronously requests more samples from the capture engine, runs ML inference on the samples (eg. Background blur or style transfer) on parallel threads as scheduled by the MF async work queue, and asynchronously signals to the capture preview sink whenever a sample has been processed. --> ```BackgroundBlur/TransformAsync.h```
3. Defines the IStreamModel interface, which is implemented by the StyleTransfer and BackgroundBlur models. This interface manages cached copies of the input/output sample video frames, maintains a LearningModelSession and Binding, and executes an evaluation. --> ```BackgroundBlur/SegmentModel.h```

TODO: More information on each of these components and data flow will be included here in a separate document once it's complete.


Samples/BackgroundBlur/BackgroundBlur/Capture.h
```diff
@@ -0,0 +1,238 @@
...
+    static HRESULT CreateInstance(HWND hwndEvent, CaptureManager** ppEngine)
+    {
+        HRESULT hr = S_OK;
+        *ppEngine = NULL;
+
+        CaptureManager* pEngine = new (std::nothrow) CaptureManager(hwndEvent);
+        if (pEngine == NULL)
+        {
+            hr = E_OUTOFMEMORY;
+            goto Exit;
```
JingqiaoFu, line 171: In general, we should avoid using goto as much as possible as a good coding practice, the logic here seems easy enough to replace it with some if else statements

Samples/BackgroundBlur/BackgroundBlur/Helpers/CSampleQueue.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+        AddRef();
+    }while(false);
+
+    return hr;
+}
+
+ULONG CSampleQueue::Release(void)
+{
+    ULONG   ulRef = 0;
```
JingqiaoFu, line 93: Nit: The spacing is inconsistent in different places, I suggest use single space to separate symbols across the file.

Samples/BackgroundBlur/BackgroundBlur/Capture.h
JingqiaoFu: nvm, just noticed that this is open source code, might be a good idea to put open source code in a separate folder called 'external' or something

JingqiaoFu: Yeah I think do{} while(false) is actually a neat trick to avoid complicated nested if else statements (or duplicated return statements) and it can also be used as a way to encapsulate a few lines of code into a macro. goto on the other hand can be messy if not careful.

Samples/BackgroundBlur/BackgroundBlur/TransformAsync_IMFAsyncCallback.cpp
```diff
@@ -0,0 +1,81 @@
...
+    {
+        if ((pdwFlags == NULL) || (pdwQueue == NULL))
+        {
+            hr = E_POINTER;
+            break;
+        }
+
+        (*pdwFlags) = 0;
+        (*pdwQueue) = MFASYNC_CALLBACK_QUEUE_MULTITHREADED;
+    } while (false);
```
JingqiaoFu, line 19: Nit: If there is only one break point, maybe remove do-while statement to make it cleaner?

Samples/BackgroundBlur/BackgroundBlur/Helpers/CSampleQueue.cpp
```diff
@@ -0,0 +1,296 @@
...
+            pCurrNode = m_pHead;
+
+            m_pHead = m_pHead->pNext;
+
+            delete pCurrNode;
+        }
+
+        m_pTail = NULL;
+        m_length = 0;
+    }while(false);
```
JingqiaoFu, line 247: Probably don't need do-while here?

Samples/BackgroundBlur/BackgroundBlur/Helpers/CSampleQueue.cpp
```diff
@@ -0,0 +1,296 @@
...
+{
+    HRESULT hr = S_OK;
+
+    do
+    {
+        AutoLock lock(m_critSec);
+
+        m_pulMarkerID   = pulID;
+        m_bAddMarker    = TRUE;
+    }while(false);
```
JingqiaoFu, line 263: Don't need do-while here?

Samples/BackgroundBlur/BackgroundBlur/Helpers/CSampleQueue.cpp
```diff
@@ -0,0 +1,296 @@
...
+    m_pulMarkerID   = 0;
+
+    //InitializeCriticalSection(&m_csLock);
+}
+
+CSampleQueue::~CSampleQueue(void)
+{
+    RemoveAllSamples();
+    m_critSec.~CritSec();
+    //DeleteCriticalSection(&m_csLock);
```
JingqiaoFu, line 295: Remove comments?

Samples/BackgroundBlur/BackgroundBlur/Helpers/CSampleQueue.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+CSampleQueue::CSampleQueue(void)
+{
+    m_ulRef         = 1;
+    m_pHead         = NULL;
+    m_pTail         = NULL;
+    m_bAddMarker    = FALSE;
+    m_pulMarkerID   = 0;
+
+    //InitializeCriticalSection(&m_csLock);
```
JingqiaoFu, line 288: Remove comments?

Samples/BackgroundBlur/BackgroundBlur/Helpers/CSampleQueue.cpp
```diff
@@ -0,0 +1,296 @@
...
+}
+
+CSampleQueue::~CSampleQueue(void)
+{
+    RemoveAllSamples();
+    m_critSec.~CritSec();
```
tiagoshibata, line 294: Are you calling the destructor twice? Why?

Samples/BackgroundBlur/BackgroundBlur/Helpers/common.h
```diff
@@ -0,0 +1,40 @@
...
+    if (p)
+    {
+        p->Release();
+        p = NULL;
+    }
+}
+#endif
+
+#ifndef SAFE_DELETE
+#define SAFE_DELETE(x) if (x) { delete x; x = NULL; }
```
tiagoshibata, line 26: `delete nullptr` is a no-op (it's safe to delete null and the `if` is unneeded)

Samples/BackgroundBlur/.gitignore
```diff
@@ -0,0 +1 @@
+*.onnx
```
martinb35, line 1: do you actually want to ignore onnx files? or does this mean you aren't checking in a model?

Samples/WinMLSamplesGallery/WinMLSamplesGallery/Samples/StreamEffect/StreamEffect.xaml.cs
```diff
@@ -0,0 +1,74 @@
...
+        static extern IntPtr GetForegroundWindow();
+
+        public StreamEffect()
+        {
+            this.InitializeComponent();
+            // TODO: Currently the WinMLSamplesGalleryNative component will load
+            // the wrong version of the Microsoft.AI.MachineLearning.dll.
+            // To work around this, we make a dummy call to the builder to
+            // ensure that the dll is loaded.
+            var builder = Microsoft.AI.MachineLearning.Experimental.LearningModelBuilder.Create(11);
```
martinb35, line 33: can you use a named constant instead of "11" so we know what it's for?

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+#include <iostream>
+#include <filesystem>
+
+using winrt::Windows::Foundation::PropertyValue;
+using winrt::hstring;
+using namespace winrt;
+using namespace Windows::Foundation::Collections;
+
+enum OnnxDataType : long {
```
martinb35, line 11: is this defined in an onnx header somewhere that we can include?

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
```diff
@@ -0,0 +1,285 @@
...
+	ONNX_COMPLEX128 = 15,    // complex with float64 real and imaginary components
+
+	// Non-IEEE floating-point format based on IEEE754 single-precision
+	// floating-point number truncated to 16 bits.
+	// This format has 1 sign bit, 8 exponent bits, and 7 mantissa bits.
+	ONNX_BFLOAT16 = 16,
+}OnnxDataType;
+
+
+int g_scale = 5;
```
martinb35, line 41: I think you can make this a static const where its used instead of defining it globally.

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
```diff
@@ -0,0 +1,285 @@
...
+auto outputBindProperties = PropertySet();
```
martinb35, line 42: can you remove this? it looks unused (it gets overwritten by local scope var with same name on line 65)

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
```diff
@@ -0,0 +1,285 @@
...
+	m_binding.Bind(outputName, m_outputVideoFrame, outputBindProperties); // TODO: See if can bind videoframe from MFT
+	auto results = m_session.Evaluate(m_binding, L"");
+
+	m_outputVideoFrame.CopyToAsync(outVideoFrame).get();
+
+	m_bSyncStarted = FALSE;
+}
+LearningModel StyleTransfer::GetModel()
+{
+	auto rel = std::filesystem::current_path();
```
martinb35, line 77: name: use "model_path" instead of "rel"?

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
```diff
@@ -0,0 +1,285 @@
...
+	rel.append("Assets\\mosaic.onnx");
+	return LearningModel::LoadFromFilePath(rel + L"");
```
martinb35, line 79: why do you add L""? is that to cast the string to a different type? seems like an odd way to do that.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Capture.cpp
```diff
@@ -0,0 +1,844 @@
...
+    *uiEncodingBitrate = (UINT32) uiBitrate;
+
+done:
+
+    return hr;
+}
+
+HRESULT ConfigureVideoEncoding(IMFCaptureSource *pSource, IMFCaptureRecordSink *pRecord, REFGUID guidEncodingType)
+{
+    IMFMediaType *pMediaType = NULL;
```
chrilaMSFT, line 531: >IMFMediaType [](http://example.com/codeflow?start=4&length=12)

I think using WIL auto pointers and return logic for external samples, it can remove issues with leaks and make the code smaller

Samples/BackgroundBlur/BackgroundBlur/Helpers/CSampleQueue.h
```diff
@@ -0,0 +1,67 @@
+#pragma once
+
+#include <windows.h>
+#include <Mfidl.h>
+#include "common.h"
+#include <mutex> 
+
+using namespace MediaFoundationSamples;
+
+class CSampleQueue: public IUnknown
```
chrilaMSFT, line 10: >CSampleQueue [](http://example.com/codeflow?start=6&length=12)

Why not just use stl::list?

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
```diff
@@ -0,0 +1,285 @@
...
+	ONNX_INT32 = 6,
+	ONNX_INT64 = 7,
+	ONNX_STRING = 8,
+	ONNX_BOOL = 9,
+
+	// IEEE754 half-precision floating-point format (16 bits wide).
+	// This format has 1 sign bit, 5 exponent bits, and 10 mantissa bits.
+	ONNX_FLOAT16 = 10,
+
+	ONNX_DOUBLE = 11,
```
chrilaMSFT, line 28: >	 [](http://example.com/codeflow?start=0&length=1)

NIT:  Most of this code is using tabs not spaces

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.h
```diff
@@ -0,0 +1,171 @@
...
+public:
+	BackgroundBlur(int w, int h) : 
+		IStreamModel(w, h)
+	{
+		SetModels(w, h);
+	}
+	BackgroundBlur() : 
+		IStreamModel()
+	{};
+	~BackgroundBlur();
```
chrilaMSFT, line 158: I think you need virtual destruct

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.h
```diff
@@ -0,0 +1,171 @@
...
+using namespace winrt::Microsoft::AI::MachineLearning::Experimental;
+using namespace winrt::Windows::Graphics::DirectX::Direct3D11;
+using namespace winrt::Windows::Media;
+
+// Model-agnostic helper LearningModels
+LearningModel Normalize0_1ThenZScore(long height, long width, long channels, const std::array<float, 3>& means, const std::array<float, 3>& stddev);
+LearningModel ReshapeFlatBufferToNCHW(long n, long c, long h, long w);
+LearningModel Invert(long n, long c, long h, long w);
+
+class IStreamModel
```
chrilaMSFT, line 32: >IStreamModel [](http://example.com/codeflow?start=6&length=12)

I would rename since it is more of a class CStreamModel instead of IStreamModel

Samples/BackgroundBlur/BackgroundBlur/Capture.cpp
```diff
@@ -0,0 +1,517 @@
...
+    if(SUCCEEDED(hr))
+    {
+        hr = g_pDXGIMan->ResetDevice(g_pDX11Device, g_ResetToken);
+    }
+        
+    return hr;
+}
+
+HRESULT
+CaptureManager::InitializeCaptureManager(HWND hwndPreview, IUnknown* pUnk)
```
chrilaMSFT, line 177: If this is a sample I would follow 

HRESULT PerformOperation(int operation) noexcept try
{
    vector<int> items = GetItems(operation);
    SomeFunction(items);
    return S_OK;
}
CATCH_RETURN();

instead of using on throw news, this would allow for modification in future without letting an exception leak or sub-function changing its impl

see [https://www.osgwiki.com/wiki/Cpp_Core_Guidelines#PATTERNS.1:_Follow_the_error_code_based_Try_pattern](https://www.osgwiki.com/wiki/Cpp_Core_Guidelines#PATTERNS.1:_Follow_the_error_code_based_Try_pattern)

Samples/BackgroundBlur/BackgroundBlur/TransformAsync.cpp
```diff
@@ -0,0 +1,1212 @@
...
+#include <winrt/windows.foundation.h>
+#include <windows.graphics.directx.direct3d11.interop.h>
+#include <winrt/Windows.Graphics.DirectX.Direct3D11.h>
+#include <windows.graphics.imaging.interop.h>
+#include <windows.media.core.interop.h>
+#include "Helpers/common.h"
+#include <unknwn.h>
+#include <stdio.h>
+#include <wchar.h>
+#include <future>
```
chrilaMSFT, line 12: I would clean up code to use Std header will simplify the dependencies and include logic

Samples/BackgroundBlur/BackgroundBlur/TransformAsync.h
```diff
@@ -0,0 +1,353 @@
...
+    HANDLE                          m_hDeviceHandle;        // Handle to the current device
+    // Immediate device context
+    com_ptr<ID3D11DeviceContext4>       m_spContext;
+    com_ptr<ID3D11Device5>              m_spDevice;
+    com_ptr<IMFVideoSampleAllocatorEx>  m_spOutputSampleAllocator;  // Allocates d3d-backed output samples. 
+    com_ptr<IMFVideoSampleAllocatorCallback> m_spOutputSampleCallback; // Callback for when a frame is returns to the output sample allocator. 
+    // Frame rate synch objects
+    com_ptr<ID3D11Fence>                m_spFence;
+    UINT64 m_fenceValue;
+    HANDLE m_hFrameThread;
```
chrilaMSFT, line 343: >HANDLE [](http://example.com/codeflow?start=4&length=6)

use wil::unique_handle

Samples/BackgroundBlur/BackgroundBlur/TransformAsync.cpp
```diff
@@ -0,0 +1,1212 @@
...
+    m_bShutdown = FALSE;
+    m_bFirstSample = TRUE;
+    // m_bDXVA = TRUE; // TODO: Always true to be fast
+    m_pInputSampleQueue = NULL;
+    m_pOutputSampleQueue = NULL;
+    m_videoFOURCC = 0;
+    m_imageWidthInPixels = 0; 
+    m_imageHeightInPixels = 0;
+
+    m_spDeviceManager = NULL;
```
chrilaMSFT, line 77: I would use the pattern to init all variables in header where they are declared 

example 

volatile ULONG                  m_nRefCount; = 1;

Samples/BackgroundBlur/BackgroundBlur/TransformAsync.cpp
```diff
@@ -0,0 +1,1212 @@
...
+    if (m_spDeviceManager)
+    {
+        m_spDeviceManager->CloseDeviceHandle(m_hDeviceHandle);
+        m_spDeviceManager->Release();
+    }
+
+    if (m_spOutputSampleAllocator) {
+        m_spOutputSampleAllocator->UninitializeSampleAllocator();
+        //m_spOutputSampleAllocator->Release();
+    }
```
chrilaMSFT, line 104: Should this be done in shutdown instead?

Samples/BackgroundBlur/BackgroundBlur/TransformAsync.cpp
```diff
@@ -0,0 +1,1212 @@
...
+        return 0;
+    }
+    return 1;
+}
+
+HRESULT TransformAsync::OnSetD3DManager(ULONG_PTR ulParam)
+{
+    HRESULT hr = S_OK;
+    com_ptr<IUnknown> pUnk;
+    com_ptr<IMFDXGIDeviceManager> pDXGIDeviceManager;
```
chrilaMSFT, line 377: >com_ptr [](http://example.com/codeflow?start=4&length=7)

Should you use com_ptr_nothrow?

Samples/BackgroundBlur/BackgroundBlur/TransformAsync.cpp
```diff
@@ -0,0 +1,1212 @@
...
+
+done:
+    return hr;
+}
+
+// TODO: Clean up this function
+HRESULT TransformAsync::UpdateDX11Device()
+{
+    HRESULT hr = S_OK;
+    com_ptr<ID3D11DeviceContext> pContext;
```
chrilaMSFT, line 420: >pContext [](http://example.com/codeflow?start=33&length=8)

looks like mixing name coding between sp and p prefixes to smart ptrs.  Generally I believe the guidance is to remove for external code 

[https://www.osgwiki.com/wiki/Cpp_Core_Guidelines#NL.5_Don.27t_encode_type_information_in_names](https://www.osgwiki.com/wiki/Cpp_Core_Guidelines#NL.5_Don.27t_encode_type_information_in_names)

Samples/BackgroundBlur/BackgroundBlur/TransformAsync.cpp
```diff
@@ -0,0 +1,1212 @@
...
+
+                CHECK_HR(hr = MFCreateVideoSampleAllocatorEx(IID_PPV_ARGS(&spVideoSampleAllocator)));
+                spDXGIManagerUnk = m_spDeviceManager.as<IUnknown>();
+
+                CHECK_HR(hr = spVideoSampleAllocator->SetDirectXManager(spDXGIManagerUnk.get()));
+
+                m_spOutputSampleAllocator.attach(spVideoSampleAllocator.detach());
+            }
+
+            CHECK_HR(hr = m_spOutputSampleAllocator->InitializeSampleAllocatorEx(2, m_numThreads, m_spAllocatorAttributes.get(), m_spOutputType.get()));
```
chrilaMSFT, line 485: >2, m_numThreads [](http://example.com/codeflow?start=81&length=15)

I would use defines for these

Samples/BackgroundBlur/BackgroundBlur/TransformAsync.h
```diff
@@ -0,0 +1,353 @@
...
+    // Frame rate synch objects
+    com_ptr<ID3D11Fence>                m_spFence;
+    UINT64 m_fenceValue;
+    HANDLE m_hFrameThread; 
+    HANDLE m_hFenceEvent; // Handle to the fence complete event
+
+
+    // Model Inference fields
+    int m_numThreads =                                     // Number of threads running inference in parallel.
+         max(std::thread::hardware_concurrency(), 5);
```
chrilaMSFT, line 349: >5 [](http://example.com/codeflow?start=50&length=1)

why 5

Samples/BackgroundBlur/BackgroundBlur/TransformAsync.cpp
```diff
@@ -0,0 +1,1212 @@
...
+        }
+
+        /***************************************
+        ** Since this in an internal function
+        ** we know m_spEventQueue can never be
+        ** NULL due to InitializeTransform()
+        ***************************************/
+
+        {
+            std::lock_guard<std::recursive_mutex> lock(m_mutex);
```
chrilaMSFT, line 1133: why are you reaquiring the lock create event and set property should be fast so no need to unlock

Samples/BackgroundBlur/BackgroundBlur/TransformAsync_IMFAsyncCallback.cpp
```diff
@@ -0,0 +1,119 @@
+#include "TransformAsync.h"
+#include <format>
+
+int g_now;
```
chrilaMSFT, line 4: this is a redefined global

long long g_now; in TransformAsync.cpp

chrilaMSFT, line 4: >g_now [](http://example.com/codeflow?start=4&length=5)

 I would avoid globals variables if possible since there could be multiple instances of the class

Samples/BackgroundBlur/BackgroundBlur/winmain.cpp
```diff
@@ -0,0 +1,572 @@
...
+HPOWERNOTIFY    g_hPowerNotifyMonitor = NULL;
+SYSTEM_POWER_CAPABILITIES   g_pwrCaps{};
+bool            g_fSleepState = false;
+
+INT_PTR CALLBACK ChooseDeviceDlgProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lParam);
+
+#include <filesystem>
+#include <shlobj.h>
+
+static std::wstring GetLatestWinPixGpuCapturerPath_Cpp17()
```
chrilaMSFT, line 23: >static [](http://example.com/codeflow?start=0&length=6)

Break DLL effect into different folder from the main code so it can be reused independently

Samples/BackgroundBlur/.gitignore
linnealovespie: I want to ignore onnx files by default, because FCN-Resnet is a big file and I download it if needed, as defined in BackgroundBlur.vcxproj. I do check in mosaic.onnx to BackgroundBlur/Assets because it's a slightly older version of the style transfer model than is checked into model zoo, which has free dimensions and now would require some preprocessing steps.

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
linnealovespie: removed

Samples/BackgroundBlur/BackgroundBlur/Helpers/CSampleQueue.h
linnealovespie: Mainly because CSampleQueue came from demo code and because it's thread-safe without having to add lock guards around access to stl::list.

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
linnealovespie: I got this enum from the Tensorization sample in the gallery [here](https://github.com/microsoft/Windows-Machine-Learning/blob/user/linneamay/stream-async/Samples/WinMLSamplesGallery/WinMLSamplesGallery/Samples/TensorizationModels.cs), it looks like it's not defined in any onnx header but @smk2007 can probably confirm.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Capture.cpp
linnealovespie: Removed the lingering raw pointers in capture

Samples/BackgroundBlur/BackgroundBlur/TransformAsync_IMFAsyncCallback.cpp
linnealovespie: This global is separate from the TransformAsync, it's used only by FrameThreadProc to measure the time between calls to the allocator's NotifyRelease.

Samples/BackgroundBlur/BackgroundBlur/SegmentModel.cpp
smk2007: Its part of the onnx protobuf... not easily accessible from c#

PR #472, opened 2022-04-14
Add event trace for CPU fallback

JingqiaoFu:
Add event trace for CPU fallback.

The output warnings for fall back include op name, op type and duration in us

image

### DIFF
--- a/Tools/WinMLRunner/src/EventTraceHelper.cpp
+++ b/Tools/WinMLRunner/src/EventTraceHelper.cpp
@@ -202,7 +202,7 @@ static DWORD GetEventInformation(PEVENT_RECORD pEvent, PTRACE_EVENT_INFO& pInfo)
     return status;
 }
 
-static std::wstring& TrimOperatorName(std::wstring& s)
+static std::wstring TrimOperatorName(std::wstring& s)
 {
     wstring substring = L"_kernel_time";
     int offset = (int)s.find(substring);
--- a/Tools/WinMLRunner/src/EventTraceHelper.cpp
+++ b/Tools/WinMLRunner/src/EventTraceHelper.cpp
@@ -1,5 +1,5 @@
-#include "EventTraceHelper.h"
 #include <direct.h>
+#include "EventTraceHelper.h"
 
 #define LOGSESSION_NAME L"winml"
 
--- a/Tools/WinMLRunner/src/EventTraceHelper.cpp
+++ b/Tools/WinMLRunner/src/EventTraceHelper.cpp
@@ -24,7 +24,7 @@ Level Verbosity is as follows:
 From <https://docs.microsoft.com/en-us/message-analyzer/system-etw-provider-event-keyword-level-settings>  
 */
 
-void EventTraceHelper::RemoveTrailingSpace(PEVENT_MAP_INFO pMapInfo)
+static void RemoveTrailingSpace(PEVENT_MAP_INFO pMapInfo)
 {
     DWORD byteLength = 0;
 
@@ -35,7 +35,7 @@ void EventTraceHelper::RemoveTrailingSpace(PEVENT_MAP_INFO pMapInfo)
     }
 }
 
-DWORD EventTraceHelper::GetArraySize(PEVENT_RECORD pEvent, PTRACE_EVENT_INFO pInfo, USHORT i, PUSHORT ArraySize)
+static DWORD GetArraySize(PEVENT_RECORD pEvent, PTRACE_EVENT_INFO pInfo, USHORT i, PUSHORT ArraySize)
 {
     DWORD status = ERROR_SUCCESS;
     PROPERTY_DATA_DESCRIPTOR dataDescriptor;
@@ -64,7 +64,7 @@ DWORD EventTraceHelper::GetArraySize(PEVENT_RECORD pEvent, PTRACE_EVENT_INFO pIn
 // map values can be integer values or bit values. If the property specifies a value
 // map, get the map.
 
-DWORD EventTraceHelper::GetMapInfo(PEVENT_RECORD pEvent, LPWSTR pMapName, DWORD DecodingSource,
+static DWORD GetMapInfo(PEVENT_RECORD pEvent, LPWSTR pMapName, DWORD DecodingSource,
                                    PEVENT_MAP_INFO& pMapInfo)
 {
     DWORD status = ERROR_SUCCESS;
@@ -111,7 +111,7 @@ DWORD EventTraceHelper::GetMapInfo(PEVENT_RECORD pEvent, LPWSTR pMapName, DWORD 
     return status;
 }
 
-DWORD EventTraceHelper::GetPropertyLength(PEVENT_RECORD pEvent, PTRACE_EVENT_INFO pInfo, USHORT i, PUSHORT PropertyLength)
+static DWORD GetPropertyLength(PEVENT_RECORD pEvent, PTRACE_EVENT_INFO pInfo, USHORT i, PUSHORT PropertyLength)
 {
     DWORD status = ERROR_SUCCESS;
     PROPERTY_DATA_DESCRIPTOR dataDescriptor;
@@ -170,7 +170,7 @@ DWORD EventTraceHelper::GetPropertyLength(PEVENT_RECORD pEvent, PTRACE_EVENT_INF
     return status;
 }
 
-DWORD EventTraceHelper::GetEventInformation(PEVENT_RECORD pEvent, PTRACE_EVENT_INFO& pInfo)
+static DWORD GetEventInformation(PEVENT_RECORD pEvent, PTRACE_EVENT_INFO& pInfo)
 {
     DWORD status = ERROR_SUCCESS;
     DWORD bufferSize = 0;
@@ -202,7 +202,7 @@ DWORD EventTraceHelper::GetEventInformation(PEVENT_RECORD pEvent, PTRACE_EVENT_I
     return status;
 }
 
-std::wstring& EventTraceHelper::TrimOperatorName(std::wstring& s)
+static std::wstring& TrimOperatorName(std::wstring& s)
 {
     wstring substring = L"_kernel_time";
     int offset = (int)s.find(substring);
@@ -232,7 +232,7 @@ std::wstring& EventTraceHelper::TrimOperatorName(std::wstring& s)
 /// \return A DWORD with a windows error value. If the function succeded, it returns
 ///     ERROR_SUCCESS.
 ///
-DWORD EventTraceHelper::GetFormatedPropertyData(_In_ const PEVENT_RECORD EventRecord,
+static DWORD GetFormatedPropertyData(_In_ const PEVENT_RECORD EventRecord,
                                               _In_ const PTRACE_EVENT_INFO EventInfo,
                               _In_ USHORT Index, _Inout_ PBYTE& UserData, _In_ PBYTE EndOfUserData,
                               _Inout_ wstring& propertyValue)
@@ -353,7 +353,7 @@ DWORD EventTraceHelper::GetFormatedPropertyData(_In_ const PEVENT_RECORD EventRe
 /// \return A DWORD with a windows error value. If the function succeded, it returns
 ///     ERROR_SUCCESS.
 ///
-DWORD EventTraceHelper::FormatDataForCPUFallback(_In_ const PEVENT_RECORD EventRecord,
+static DWORD FormatDataForCPUFallback(_In_ const PEVENT_RECORD EventRecord,
                                                _In_ const PTRACE_EVENT_INFO EventInfo)
 {
     DWORD status = ERROR_SUCCESS;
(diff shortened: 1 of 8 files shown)


Tools/WinMLRunner/src/Run.cpp
```diff
@@ -2,6 +2,7 @@
 #include "Common.h"
 #include "OutputHelper.h"
 #include "BindingUtilities.h"
+#include "EventTraceHelper.h"
```
smk2007, line 5: nit: alpha

Tools/WinMLRunner/src/Run.cpp
```diff
@@ -561,6 +563,7 @@ int run(CommandLineArgs& args,
             ConcurrentLoadModel(modelPaths, args.NumThreads(), args.ThreadInterval(), true);
             return 0;
         }
+        traceHelper.Start();
```
smk2007, line 566: >traceHelper [](http://example.com/codeflow?start=8&length=11)

This is only enabled for non-concurrent model load. Can we add a message to indicate that it is not going to actualy trace in this configuration?

Tools/WinMLRunner/src/EventTraceHelper.h
```diff
@@ -0,0 +1,49 @@
...
+
+class EventTraceHelper
+{
+private:
+    TRACEHANDLE traceHandle;
+    TRACEHANDLE sessionHandle;
+    PEVENT_TRACE_PROPERTIES sessionProperties;
+    ThreadPool threadPool;
+    static DWORD PointerSize;
+    static CommandLineArgs commandArgs;
```
smk2007, line 22: >CommandLineArgs [](http://example.com/codeflow?start=11&length=15)

Why is this static? Shouldnt it be an instance variable?

Tools/WinMLRunner/src/EventTraceHelper.h
```diff
@@ -0,0 +1,49 @@
...
+using namespace std;
+
+class EventTraceHelper
+{
+private:
+    TRACEHANDLE traceHandle;
+    TRACEHANDLE sessionHandle;
+    PEVENT_TRACE_PROPERTIES sessionProperties;
+    ThreadPool threadPool;
+    static DWORD PointerSize;
```
smk2007, line 21: >PointerSize [](http://example.com/codeflow?start=17&length=11)

Can this be moved into the cpp file to not pollute the header interface file?

Tools/WinMLRunner/src/EventTraceHelper.h
```diff
@@ -0,0 +1,49 @@
...
+    static void RemoveTrailingSpace(PEVENT_MAP_INFO pMapInfo);
+    static DWORD GetArraySize(PEVENT_RECORD pEvent, PTRACE_EVENT_INFO pInfo, USHORT i, PUSHORT ArraySize);
+    static DWORD GetMapInfo(PEVENT_RECORD pEvent, LPWSTR pMapName, DWORD DecodingSource, PEVENT_MAP_INFO& pMapInfo);
+    static DWORD GetPropertyLength(PEVENT_RECORD pEvent, PTRACE_EVENT_INFO pInfo, USHORT i, PUSHORT PropertyLength);
+    static DWORD GetEventInformation(PEVENT_RECORD pEvent, PTRACE_EVENT_INFO& pInfo);
+    static wstring& TrimOperatorName(wstring& s);
+    static DWORD GetFormatedPropertyData(_In_ const PEVENT_RECORD EventRecord, _In_ const PTRACE_EVENT_INFO EventInfo,
+                                         _In_ USHORT Index, _Inout_ PBYTE& UserData, _In_ PBYTE EndOfUserData,
+                                         _Inout_ wstring& propertyValue);
+    static DWORD FormatDataForCPUFallback(_In_ const PEVENT_RECORD EventRecord, _In_ const PTRACE_EVENT_INFO EventInfo);
```
smk2007, line 47: Do these APIs need to be in the header, or can they be static methods in the cpp file?

Tools/WinMLRunner/src/EventTraceHelper.cpp
```diff
@@ -0,0 +1,559 @@
+#include "EventTraceHelper.h"
+#include <direct.h>
+
+#define LOGSESSION_NAME _T("winml")
```
smk2007, line 4: >_T [](http://example.com/codeflow?start=24&length=2)

_T macro was useful in the old days when unicode wasnt supported, why cant winml just be whar_t?

Tools/WinMLRunner/src/EventTraceHelper.h
```diff
@@ -0,0 +1,49 @@
...
+#include "ThreadPool.h"
+#include <in6addr.h>
+#include "CommandLineArgs.h"
+
+using namespace std;
+
+class EventTraceHelper
+{
+private:
+    TRACEHANDLE traceHandle;
```
smk2007, line 17: >traceHandle [](http://example.com/codeflow?start=16&length=11)

It is nice to have member var prefixed with _ so that the members are clearly visible in the implementation.

Tools/WinMLRunner/src/EventTraceHelper.cpp
smk2007: Later on you assume it is wchar_t via sizeof(wchar_t)

Tools/WinMLRunner/src/EventTraceHelper.cpp
```diff
@@ -0,0 +1,559 @@
...
+    catch (exception ex)
+    {
+        printf("Exception when processing event trace: %s\n", ex.what());
+        // catch exceptions occurred in event_callback
+    }
+}
+
+void EventTraceHelper::Start()
+{
+    int bufferSize = sizeof(EVENT_TRACE_PROPERTIES) + (sizeof(LOGSESSION_NAME) + 1) * sizeof(wchar_t);
```
smk2007, line 502: >+ (sizeof(LOGSESSION_NAME) + 1) * sizeof(wchar_t); [](http://example.com/codeflow?start=52&length=50)

It looks like the start trace function describes this weird memory layout... can you add a link explaining why this extra memory is being added to the struct:

[https://docs.microsoft.com/en-us/windows/win32/api/evntrace/ns-evntrace-event_trace_properties](https://docs.microsoft.com/en-us/windows/win32/api/evntrace/ns-evntrace-event_trace_properties)

Tools/WinMLRunner/src/EventTraceHelper.cpp
```diff
@@ -0,0 +1,559 @@
...
+    sessionProperties->Wnode.Guid = guid;
+    sessionProperties->Wnode.ClientContext = 0;
+    sessionProperties->Wnode.Flags = WNODE_FLAG_TRACED_GUID;
+    sessionProperties->LogFileMode = EVENT_TRACE_REAL_TIME_MODE;
+    sessionProperties->LoggerNameOffset = sizeof(EVENT_TRACE_PROPERTIES);
+
+    auto hr = StartTrace(static_cast<PTRACEHANDLE>(&sessionHandle), LOGSESSION_NAME, sessionProperties);
+    if (hr != ERROR_SUCCESS)
+    {
+        printf("Error starting event trace: Trace already started %s\n", GetLastError());
```
smk2007, line 518: Is this an error or a warning?

If it is an error, can we return; immediately?
Otherwise, can we say its a warning?

Tools/WinMLRunner/src/EventTraceHelper.cpp
```diff
@@ -0,0 +1,559 @@
+#include "EventTraceHelper.h"
+#include <direct.h>
+
+#define LOGSESSION_NAME _T("winml")
+
+struct __declspec(uuid("{BCAD6AEE-C08D-4F66-828C-4C43461A033D}")) DXGKRNL_PROVIDER_GUID_HOLDER;
```
smk2007, line 6: >DXGKRNL_PROVIDER_GUID_HOLDER [](http://example.com/codeflow?start=66&length=28)

This should be WINML_PROVIDER_GUID_HOLDER right?

Tools/WinMLRunner/src/EventTraceHelper.cpp
```diff
@@ -0,0 +1,559 @@
+#include "EventTraceHelper.h"
+#include <direct.h>
+
+#define LOGSESSION_NAME _T("winml")
+
+struct __declspec(uuid("{BCAD6AEE-C08D-4F66-828C-4C43461A033D}")) DXGKRNL_PROVIDER_GUID_HOLDER;
+
+static const auto DXGKRNL_PROVIDER_GUID = __uuidof(DXGKRNL_PROVIDER_GUID_HOLDER);
```
smk2007, line 8: >DXGKRNL_PROVIDER_GUID [](http://example.com/codeflow?start=18&length=21)

WINML_PROVIDER_GUID

Tools/WinMLRunner/src/EventTraceHelper.cpp
```diff
@@ -0,0 +1,559 @@
...
+
+    traceHandle = OpenTrace(&loggerInfo);
+    if (traceHandle == INVALID_PROCESSTRACE_HANDLE)
+    {
+        printf("Error opening event trace: OpenTrace failed with %lu\n", GetLastError());
+        throw std::runtime_error("Unable to open trace");
+    }
+
+    if (pHeader->PointerSize != sizeof(PVOID))
+        pHeader = (PTRACE_LOGFILE_HEADER)((PUCHAR)pHeader + 2 * (pHeader->PointerSize - sizeof(PVOID)));
```
smk2007, line 547: This seems to update pHeader, which is a pointer into loggerInfo, which is used to open the trace.

Why do we update this? From inspeaction it looks like this will do nothing?

Tools/WinMLRunner/src/EventTraceHelper.cpp
```diff
@@ -0,0 +1,559 @@
...
+    }
+
+    return ERROR_SUCCESS;
+}
+
+VOID WINAPI EventTraceHelper::EventRecordCallback(EVENT_RECORD* pEventRecord)
+{
+    // This is where you would get the details of ETW event
+    DWORD status = ERROR_SUCCESS;
+    PTRACE_EVENT_INFO pInfo = NULL;
```
smk2007, line 428: can we make this a smart unique_ptr and it should get rid of the goto

Tools/WinMLRunner/src/EventTraceHelper.cpp
fdwr: `pHeader` - ancient Hungarian... weird :b.

Tools/WinMLRunner/src/EventTraceHelper.cpp
JingqiaoFu: Yeah a warning sounds more appropriate

Tools/WinMLRunner/src/EventTraceHelper.cpp
JingqiaoFu: Yes

Tools/WinMLRunner/src/EventTraceHelper.cpp
JingqiaoFu: Yeah will remove

Tools/WinMLRunner/src/EventTraceHelper.cpp
JingqiaoFu: Sure will fix

Tools/WinMLRunner/src/Run.cpp
JingqiaoFu: Sorry can you elaborate?

Tools/WinMLRunner/src/EventTraceHelper.h
JingqiaoFu: Sure

Tools/WinMLRunner/src/EventTraceHelper.h
JingqiaoFu: They can be static methods in cpp file, I wasn't sure if making them static member methods is better or not

Tools/WinMLRunner/src/EventTraceHelper.cpp
JingqiaoFu: decided to just restructure the code to get rid of goto, hope it's OK

fdwr:
This will be useful : ).

Tools/WinMLRunner/src/EventTraceHelper.cpp
```diff
@@ -0,0 +1,567 @@
...
+
+    if (ERROR_SUCCESS != status)
+    {
+        printf("TdhGetEventInformation failed with 0x%x.\n", status);
+    }
+
+    return status;
+}
+
+static std::wstring& TrimOperatorName(std::wstring& s)
```
smk2007, line 205: >TrimOperatorName [](http://example.com/codeflow?start=21&length=16)

A little odd that this helper is destructive

smk2007 (approved):

Tools/WinMLRunner/src/EventTraceHelper.cpp
JingqiaoFu: By destructive you mean it modifies the variable? this is a helper method that modifies an intermediate variable, should be fine?

fdwr: He just means that it was odd to return a reference to the modified string rather than a new instance. Typically a function either returns a new string or modifies it in place but returns nothing. e.g. either a or b

a. `void TrimOperatorName(std:;string& s)`
b. `std::string TrimOperatorName(std:;string_view s)`

JingqiaoFu: hmmmm, modified the method to return a new instance instead of reference

PR #480, opened 2022-06-01
Add sample to showcase binding and inferencing on DX resources using ORT

nums11:
This PR adds a sample that showcases how to bind and inference on DX resources using ORT.

3 images (cat, dog, hummingbird) are repeatedly displayed in an HWND and their corresponding resource is passed to ORT for evaluation using EfficentNet.

Important Files to look at:
- DXResourceBinding.cpp
  - Main implementation file

- ORTHelpers.cpp
  - Handles all ORT preprocessing and evaluation
  - CreateTensorValueFromD3DResource function demonstrates the main functionality of the sample.

- D3D12Quad.cpp
  - Handles all D3D-related functionality
  - CopyTextureIntoCurrentBuffer is the conversion from a texture 2d into a resource dimension buffer


Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/DXResourceBinding.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+        // Detach the thread so it doesn't block the UI
+        d3d_th.detach();
+	}
+
+    // Get the buffer currently being drawn to the screen then
+    // preprocess and classify it
+    winrt::com_array<float> DXResourceBinding::EvalORT() {
+        // Get the buffer currently being drawn to the screen
+        ComPtr<ID3D12Resource> currentBuffer = sample.GetCurrentBuffer();
```
martinb35, line 41: while running this locally, "currentBuffer" was null and crashed when passing that value into ORT. Looks like this code should handle that case.

nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGallery/Samples/DXResourceBindingORT/DXResourceBindingORT.xaml.cs
```diff
@@ -0,0 +1,108 @@
...
+            await Task.Run(() => WinMLSamplesGalleryNative.DXResourceBinding.LaunchWindow());
+
+            ToggleLoadingTxtVisibility();
+            ClassifyCube();
+        }
+
+        private async void ClassifyCube()
+        {
+            int i = 0;
+            float[] prev_results = { };
```
fdwr, line 49: `previousResults` per C# naming conventions of fully spelling out words and using camel case for locals like `pageExited` on line 52.

Samples/WinMLSamplesGallery/WinMLSamplesGallery/Samples/DXResourceBindingORT/DXResourceBindingORT.xaml.cs
```diff
@@ -0,0 +1,108 @@
...
+                }
+                float[] results = await Task.Run(() => WinMLSamplesGalleryNative.DXResourceBinding.EvalORT());
+                if (i == 0)
+                    prev_results = results;
+                // The first evaluation may return null so move to the next iteration
+                if (results == null)
+                    continue;
+                UpdateClassificationResults(prev_results);
+                prev_results = results;
+                System.Threading.Thread.Sleep(1000);
```
fdwr, line 65: Since this is already an `async` function, should this be an `await` statement rather than a sleep? Direct sleeps are bad :b.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
```diff
@@ -0,0 +1,166 @@
...
+    HMODULE hModule = NULL;
+    GetModuleHandleEx(
+        GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS,
+        (LPCTSTR)GetCurrentModule,
+        &hModule);
+    return hModule;
+}
+
+// Launch the HWND and init D3D
+int Win32Application::Run(D3D12Quad* pSample, int nCmdShow)
```
fdwr, line 21: Hungarian warts like `p` were already deprecated as far back as 2007 in SDK samples. Can we kill them for good in 2022?

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
```diff
@@ -0,0 +1,832 @@
...
+    HrException(HRESULT hr) : std::runtime_error(HrToString(hr)), m_hr(hr) {}
+    HRESULT Error() const { return m_hr; }
+private:
+    const HRESULT m_hr;
+};
+
+D3D12Quad::D3D12Quad(UINT width, UINT height, std::wstring name) :
+    m_width(width),
+    m_height(height),
+    m_title(name),
```
fdwr, line 24: Should `std::move` here to avoid the extra copy.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
```diff
@@ -0,0 +1,832 @@
...
+    fileNames = {
+        Win32Application::GetAssetPath(L"dog.jpg").c_str(),
+        Win32Application::GetAssetPath(L"cat.jpg").c_str(),
+        Win32Application::GetAssetPath(L"hummingbird.jpg").c_str()
+    };
+    fileIndex = 0;
+    updateCounter = 0;
+    LoadPipeline();
+    LoadAssets();
+    is_initialized = true;
```
fdwr, line 45: [nit] Win32 apps are predominantly camel case for locals and member fields. Even WinRT C++ apps use camel case for locals (e.g. https://github.com/microsoft/windows-appsample-photo-editor/blob/master/PhotoEditor/Photo.cpp) along with Win32, DirectX, and pretty much every other Microsoft codebase, with the rare oddball exception of ONNX Runtime (probably because the founders originated from TensorFlow) and certain classes in WinRT C++ projections (because those mimic the STL). Granted, this is a demo on using ORT, but the app itself is win32 (and it's good to be internally self-consistent with all the other variables like `updateCounter` and `fileIndex`).

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
```diff
@@ -0,0 +1,832 @@
...
+// load and decode image from file
+int D3D12Quad::LoadImageDataFromFile(BYTE** imageData, D3D12_RESOURCE_DESC& resourceDescription, LPCWSTR filename, int& bytesPerRow)
+{
+    HRESULT hr;
+
+    static IWICImagingFactory* wicFactory;
+
+    // reset decoder, frame and converter since these will be different for each image we load
+    IWICBitmapDecoder* wicDecoder = NULL;
+    IWICBitmapFrameDecode* wicFrame = NULL;
```
fdwr, line 628: Let's update to `nullptr`. I'm not concerned with compilers before 2011, as those can't build WinRT apps anyway.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/ORTHelpers.cpp
```diff
@@ -0,0 +1,157 @@
...
+    ortApi.AddFreeDimensionOverrideByName(sessionOptions, "batch_size", 1);
+    OrtSessionOptionsAppendExecutionProvider_DML(sessionOptions, 0);
+
+    return Ort::Session(ortEnvironment, model_file_path, sessionOptions);
+}
+
+
+// Run the buffer through a preprocessing model that will shrink the
+// image from 512 x 512 x 4 to 224 x 224 x 3
+Ort::Value Preprocess(Ort::Session& session,
```
fdwr, line 49: [nit] Awkward to read with strange indent. Any of these will work:

a.
```
Ort::Value Preprocess(Ort::Session& session, ComPtr<ID3D12Resource> currentBuffer)
```

b.
```
Ort::Value Preprocess(
    Ort::Session& session,
    ComPtr<ID3D12Resource> currentBuffer
)
```

c.
```
Ort::Value Preprocess(Ort::Session& session,
                      ComPtr<ID3D12Resource> currentBuffer)
```

Appears (b) is the most self-consistent with surrounding code.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/ORTHelpers.cpp
```diff
@@ -0,0 +1,157 @@
...
+    ComPtr<ID3D12Resource> currentBuffer)
```
fdwr, line 50: We should just pass `ID3D12Resource*` since we don't need the actual `ComPtr` and should be able to accept any COM pointer or even raw pointer, and passing by value just increments the ComPtr unnecessarily (atomic refs).

Plus, "be conservative in what you do, be liberal in what you accept from others".
https://en.wikipedia.org/wiki/Robustness_principle

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
```diff
@@ -0,0 +1,166 @@
...
+    {
+        // Destroy the sample and window if the close_window
+        // variable is set to true
+        if (close_window) {
+            pSample->OnDestroy();
+            PostMessage(m_hwnd, WM_CLOSE, 0, 0);
+            break;
+        }
+        // Process any messages in the queue.
+        if (PeekMessage(&msg, NULL, 0, 0, PM_REMOVE))
```
fdwr, line 85: Use `GetMessage` to wait for the next event. `PeekMessage` burns a bunch of extra CPU cycles because Peek returns immediately if there are no messages, making this a tight loop. It's useful in specific scenarios where you are interleaving other work or where you are constantly displaying frames as fast as possible (many games).

```
    MSG msg;

    // Main message loop:
    while (GetMessage(&msg, nullptr, 0, 0))
    {
        TranslateMessage(&msg);
        DispatchMessage(&msg);
    }
```

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
```diff
@@ -0,0 +1,166 @@
...
+    // the sample is reloaded
+    if (!UnregisterClass(L"DXQuad", hInstance))
+    {
+        auto error = GetLastError();
+        MessageBox(NULL, L"Error unregistering class",
+            L"Error", MB_OK | MB_ICONERROR);
+        return 1;
+    }
+    else {
+        OutputDebugString(L"Class unregistered");
```
fdwr, line 102: We shouldn't need debug output in the success case.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+// Main message handler for the sample.
+LRESULT CALLBACK Win32Application::WindowProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
+{
+    D3D12Quad* pSample = reinterpret_cast<D3D12Quad*>(GetWindowLongPtr(hWnd, GWLP_USERDATA));
+
+    switch (message)
+    {
+    case WM_CREATE:
+    {
```
fdwr, line 117: [nit] Lines 125 to 133 indent.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+void Win32Application::CloseWindow()
+{
+    close_window = true;
+}
+
+std::wstring Win32Application::GetAssetPath(LPCWSTR assetName)
+{
+    WCHAR modulePath[512];
+    UINT modulePathSize = _countof(modulePath);
```
fdwr, line 152: Can we use `std::size`? `_countof` is MSVC specific.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/DXResourceBinding.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+Microsoft::WRL::ComPtr<ID3D12Resource> d3dResource;
+static std::optional<Ort::Session> preprocesingSession;
+static std::optional<Ort::Session> inferenceSession;
+D3D12Quad sample(800, 600, L"D3D12 Quad");
+
+namespace winrt::WinMLSamplesGalleryNative::implementation
+{
+    // Create ORT Sessions and launch D3D window in a separate thread
+	void DXResourceBinding::LaunchWindow() {
```
fdwr, line 22: Odd indent.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
```diff
@@ -0,0 +1,166 @@
...
+    ShowWindow(m_hwnd, nCmdShow);
+
+    // Main sample loop.
+    close_window = false;
+    MSG msg = {};
+    while (msg.message != WM_QUIT)
+    {
+        // Destroy the sample and window if the close_window
+        // variable is set to true
+        if (close_window) {
```
fdwr, line 79: >  [](http://example.com/codeflow?start=25&length=1)

[nit] consistent brace alignment

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
```diff
@@ -0,0 +1,166 @@
...
+        windowRect.right - windowRect.left,
+        windowRect.bottom - windowRect.top,
+        nullptr,        // We have no parent window.
+        nullptr,        // We aren't using menus.
+        hInstance,
+        pSample);
+    if (!m_hwnd)
+    {
+        MessageBox(NULL, L"Error creating window",
+            L"Error", MB_OK | MB_ICONERROR);
```
fdwr, line 64: [nit] Easier to read to just put on a single line instead of a weird split. 80 character line wrap is an ancient requirement from the 1990's that don't apply to today's 1080p above monitors. Plus for consistency, other nearby lines are even longer, and yet we don't wrap those.

```
    RECT windowRect = { 0, 0, static_cast<LONG>(pSample->GetWidth()), static_cast<LONG>(pSample->GetHeight()) };
LRESULT CALLBACK Win32Application::WindowProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
```

Samples/WinMLSamplesGallery/WinMLSamplesGallery/Pages/SampleBasePage.xaml.cs
```diff
@@ -54,6 +57,11 @@ protected override async void OnNavigatingFrom(NavigatingCancelEventArgs e)
             {
                 var page = (Samples.Batching)SampleFrame.Content;
                 page.StopAllEvents();
+            } else if(SampleFrame.SourcePageType == typeof(Samples.DXResourceBindingORT))
+            {
+                var page = (Samples.DXResourceBindingORT)SampleFrame.Content;
+                page.StopAllEvents();
+
             }
```
fdwr, line 65: [nit] extra line. Propose moving this line break to line 60 instead :b.

}
else if (...)
{

Samples/WinMLSamplesGallery/WinMLSamplesGallery/Samples/DXResourceBindingORT/DXResourceBindingORT.xaml.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            index = i;
+            label = lab;
+        }
+    }
+
+    public sealed partial class DXResourceBindingORT : Page
+    {
+        private bool pageExited = false;
+        public DXResourceBindingORT()
```
fdwr, line 30: [nit] Blank line after class variables before functions.

Samples/WinMLSamplesGallery/WinMLSamplesGallery/Samples/DXResourceBindingORT/DXResourceBindingORT.xaml.cs
```diff
@@ -0,0 +1,108 @@
...
+        private void UpdateClassificationResults(float[] results)
+        {
+            var results_lst = new List<float>(results);
+            List<int> top_1k_imagenet_indices = Enumerable.Range(0, 1000).ToList();
+            top_1k_imagenet_indices.Sort((x, y) => results_lst[y].CompareTo(results_lst[x]));
+            // The top 10 results represented by their indices in the list of ImageNet labels
+            List<int> top_10_imagenet_indices = top_1k_imagenet_indices.Take(10).ToList();
+
+            List<DXClassifierResult> dx_classifier_results = new List<DXClassifierResult>();
+            for (int i = 0; i < 10; i++)
```
fdwr, line 81: >10 [](http://example.com/codeflow?start=32&length=2)

For robustness, take min of 10 and  top_10_imagenet_indices's length. Otherwise `top_10_imagenet_indices[i]` is out of bounds if the list is smaller (it shouldn't be, but code gets moved around, and expectations change).

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/WinMLSamplesGalleryNative.vcxproj
```diff
@@ -1,7 +1,8 @@
...
-  <Import Project="..\packages\Microsoft.AI.MachineLearning.1.9.1\build\native\Microsoft.AI.MachineLearning.props" Condition="Exists('..\packages\Microsoft.AI.MachineLearning.1.9.1\build\native\Microsoft.AI.MachineLearning.props')" />
-  <Import Project="..\packages\Microsoft.AI.DirectML.1.5.1\build\Microsoft.AI.DirectML.props" Condition="Exists('..\packages\Microsoft.AI.DirectML.1.5.1\build\Microsoft.AI.DirectML.props')" />
+  <Import Project="..\packages\Microsoft.ML.OnnxRuntime.DirectML.1.11.0\build\native\Microsoft.ML.OnnxRuntime.DirectML.props" Condition="Exists('..\packages\Microsoft.ML.OnnxRuntime.DirectML.1.11.0\build\native\Microsoft.ML.OnnxRuntime.DirectML.props')" />
+  <Import Project="..\packages\Microsoft.AI.MachineLearning.1.11.0\build\native\Microsoft.AI.MachineLearning.props" Condition="Exists('..\packages\Microsoft.AI.MachineLearning.1.11.0\build\native\Microsoft.AI.MachineLearning.props')" />
+  <Import Project="..\packages\Microsoft.AI.DirectML.1.8.0\build\Microsoft.AI.DirectML.props" Condition="Exists('..\packages\Microsoft.AI.DirectML.1.8.0\build\Microsoft.AI.DirectML.props')" />
```
fdwr, line 5: >1.8.0 [](http://example.com/codeflow?start=53&length=5)

Consider using the latest:

Microsoft.AI.DirectML.1.9.0
Microsoft.ML.OnnxRuntime.DirectML.1.12.1

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/WinMLSamplesGalleryNative.vcxproj
```diff
@@ -111,6 +112,10 @@
     <UseOpenCV>False</UseOpenCV>
     <UseOpenCV Condition="Exists('$(OpenCVLibFullPath)')">True</UseOpenCV>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <IncludePath>C:\Program Files %28x86%29\Windows Kits\10\Include\10.0.19041.0\um;C:\Program Files %28x86%29\Windows Kits\10\Include\10.0.19041.0\shared;$(IncludePath)</IncludePath>
```
fdwr, line 116: ><IncludePath>C:\Program Files %28x86%29\Windows Kits\10\Include\10.0.19041.0\um;C:\Program Files %28x86%29\Windows Kits\10\Include\10.0.19041.0\shared;$(IncludePath)</IncludePath> [](http://example.com/codeflow?start=4&length=179)

Hmm, full paths shouldn't be in the vcxproj, as these won't work on other machines

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
```diff
@@ -0,0 +1,832 @@
+#include "pch.h"
+#include "stdafx.h"
+#include "D3D12Quad.h"
+
+inline std::string HrToString(HRESULT hr)
+{
+    char s_str[64] = {};
+    sprintf_s(s_str, "HRESULT of 0x%08X", static_cast<UINT>(hr));
+    return std::string(s_str);
```
fdwr, line 9: If we're using C++20 for this project (noticing "stdcpp20" in the vcxproj), then we might as well just use `std::format`.

`return std::format("HRESULT of {:08X}", static_cast<UINT>(hr))`

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
```diff
@@ -0,0 +1,832 @@
...
+
+    else if (dxgiFormat == DXGI_FORMAT_R10G10B10A2_UNORM) return 32;
+    else if (dxgiFormat == DXGI_FORMAT_B5G5R5A1_UNORM) return 16;
+    else if (dxgiFormat == DXGI_FORMAT_B5G6R5_UNORM) return 16;
+    else if (dxgiFormat == DXGI_FORMAT_R32_FLOAT) return 32;
+    else if (dxgiFormat == DXGI_FORMAT_R16_FLOAT) return 16;
+    else if (dxgiFormat == DXGI_FORMAT_R16_UNORM) return 16;
+    else if (dxgiFormat == DXGI_FORMAT_R8_UNORM) return 8;
+    else if (dxgiFormat == DXGI_FORMAT_A8_UNORM) return 8;
+}
```
fdwr, line 617: `switch` statements beat else-if chains, cleaner to read and more performant.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
```diff
@@ -0,0 +1,832 @@
...
+    else if (wicFormatGUID == GUID_WICPixelFormat32bppRGB) return GUID_WICPixelFormat32bppRGBA;
+    else if (wicFormatGUID == GUID_WICPixelFormat64bppRGB) return GUID_WICPixelFormat64bppRGBA;
+    else if (wicFormatGUID == GUID_WICPixelFormat64bppPRGBAHalf) return GUID_WICPixelFormat64bppRGBAHalf;
+#endif
+
+    else return GUID_WICPixelFormatDontCare;
+}
+
+// get the number of bits per pixel for a dxgi format
+int D3D12Quad::GetDXGIFormatBitsPerPixel(DXGI_FORMAT& dxgiFormat)
```
fdwr, line 599: >DXGI_FORMAT& dxgiFormat [](http://example.com/codeflow?start=41&length=23)

`DXGI_FORMAT dxgiFormat` pass by value - it's just a single enum.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
```diff
@@ -0,0 +1,832 @@
...
+    {
+        ThrowIfFailed(m_fence->SetEventOnCompletion(fence, m_fenceEvent));
+        WaitForSingleObject(m_fenceEvent, INFINITE);
+    }
+
+    m_frameIndex = m_swapChain->GetCurrentBackBufferIndex();
+}
+
+// get the dxgi format equivilent of a wic format
+DXGI_FORMAT D3D12Quad::GetDXGIFormatFromWICFormat(WICPixelFormatGUID& wicFormatGUID)
```
fdwr, line 527: >WICPixelFormatGUID& wicFormatGUID [](http://example.com/codeflow?start=50&length=33)

pass as constant reference (variables shouldn't be mutable unless they need to be)

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
```diff
@@ -0,0 +1,832 @@
...
+{
+    if (wicFormatGUID == GUID_WICPixelFormat128bppRGBAFloat) return DXGI_FORMAT_R32G32B32A32_FLOAT;
+    else if (wicFormatGUID == GUID_WICPixelFormat64bppRGBAHalf) return DXGI_FORMAT_R16G16B16A16_FLOAT;
+    else if (wicFormatGUID == GUID_WICPixelFormat64bppRGBA) return DXGI_FORMAT_R16G16B16A16_UNORM;
+    else if (wicFormatGUID == GUID_WICPixelFormat32bppRGBA) return DXGI_FORMAT_R8G8B8A8_UNORM;
+    else if (wicFormatGUID == GUID_WICPixelFormat32bppBGRA) return DXGI_FORMAT_B8G8R8A8_UNORM;
+    else if (wicFormatGUID == GUID_WICPixelFormat32bppBGR) return DXGI_FORMAT_B8G8R8X8_UNORM;
+    else if (wicFormatGUID == GUID_WICPixelFormat32bppRGBA1010102XR) return DXGI_FORMAT_R10G10B10_XR_BIAS_A2_UNORM;
+
+    else if (wicFormatGUID == GUID_WICPixelFormat32bppRGBA1010102) return DXGI_FORMAT_R10G10B10A2_UNORM;
```
fdwr, line 537: [nit] weird extra line

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
```diff
@@ -0,0 +1,832 @@
...
+    else if (dxgiFormat == DXGI_FORMAT_B5G6R5_UNORM) return 16;
+    else if (dxgiFormat == DXGI_FORMAT_R32_FLOAT) return 32;
+    else if (dxgiFormat == DXGI_FORMAT_R16_FLOAT) return 16;
+    else if (dxgiFormat == DXGI_FORMAT_R16_UNORM) return 16;
+    else if (dxgiFormat == DXGI_FORMAT_R8_UNORM) return 8;
+    else if (dxgiFormat == DXGI_FORMAT_A8_UNORM) return 8;
+}
+
+// load and decode image from file
+int D3D12Quad::LoadImageDataFromFile(BYTE** imageData, D3D12_RESOURCE_DESC& resourceDescription, LPCWSTR filename, int& bytesPerRow)
```
fdwr, line 620: >int [](http://example.com/codeflow?start=0&length=3)

Telling the meaning of the return type is nice for the reader. e.g.

// Load and decode image from the file, returning the number of image bytes.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.h
```diff
@@ -0,0 +1,108 @@
...
+    ComPtr<ID3D12Device> m_device;
+    ComPtr<ID3D12Resource> m_renderTargets[FrameCount];
+    ComPtr<ID3D12CommandAllocator> m_commandAllocator;
+    ComPtr<ID3D12CommandQueue> m_commandQueue;
+    ComPtr<ID3D12RootSignature> m_rootSignature;
+    ComPtr<ID3D12DescriptorHeap> m_rtvHeap;
+    ComPtr<ID3D12DescriptorHeap> m_srvHeap;
+    ComPtr<ID3D12PipelineState> m_pipelineState;
+    ComPtr<ID3D12GraphicsCommandList> m_commandList;
+    UINT m_rtvDescriptorSize;
```
fdwr, line 52: >UINT m_rtvDescriptorSize [](http://example.com/codeflow?start=4&length=24)

We should just initialize this field here (NSDMI), rather than in the constructor `D3D12Quad::D3D12Quad`, which is more brittle to change.

`UINT m_rtvDescriptorSize = 0;`

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.h
```diff
@@ -0,0 +1,108 @@
...
+    void CreateRootSignature();
+    void CreatePipelineState();
+    void CreateVertexBuffer();
+    void CreateFence();
+    void CreateDescriptorHeaps();
+    void CreateFrameResources();
+    void CreateCurrentBuffer();
+    void CopyTextureIntoCurrentBuffer();
+
+    ID3D12Resource* textureBuffer; // the resource heap containing our texture
```
fdwr, line 77: >textureBuffer [](http://example.com/codeflow?start=20&length=13)

I don't see this being initialized to nullptr here or in the constructor, nor do I see it being freed later. Leak? We should free it in the destructor (if not null, then Release), or better yet, just use WRL ComPtr. (no need to pull in all of WinRT imo, as WRL is much lighter and makes it easier to integrate into various projects for people who want to take these snippets out)

fdwr: Same with `mainDescriptorHeap` and `textureBufferUploadHeap`.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.h
```diff
@@ -0,0 +1,108 @@
...
+
+    int LoadImageDataFromFile(BYTE** imageData, D3D12_RESOURCE_DESC& resourceDescription, LPCWSTR filename, int& bytesPerRow);
```
fdwr, line 79: >int  [](http://example.com/codeflow?start=4&length=4)

`static` since it doesn't (AFAICS) use any class state.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
```diff
@@ -0,0 +1,832 @@
...
+                IID_PPV_ARGS(&adapter)));
+            ++adapterIndex)
+        {
+            DXGI_ADAPTER_DESC1 desc;
+            adapter->GetDesc1(&desc);
+
+            if (desc.Flags & DXGI_ADAPTER_FLAG_SOFTWARE)
+            {
+                // Don't select the Basic Render Driver adapter.
+                // If you want a software adapter, pass in "/warp" on the command line.
```
fdwr, line 777: >pass in "/warp" on the command line. [](http://example.com/codeflow?start=51&length=36)

Does this app accept command line parameters? I didn't see any in the main entry point.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/DXResourceBinding.cpp
```diff
@@ -0,0 +1,59 @@
...
+        // Create ORT Sessions that will be used for preprocessing and classification
+        preprocesingSession = CreateSession(Win32Application::GetAssetPath(L"dx_preprocessor_efficient_net.onnx").c_str());
+        inferenceSession = CreateSession(Win32Application::GetAssetPath(L"efficientnet-lite4-11.onnx").c_str());
+
+        // Spawn the window in a separate thread
+        std::thread d3d_th(Win32Application::Run, &sample, 10);
+
+        // Wait until the D3D pipeline finishes
+        while (!sample.is_initialized) {}
```
fdwr, line 31: >while (!sample.is_initialized) {} [](http://example.com/codeflow?start=8&length=33)

Rather than a busy wait that just eats CPU cycles, we can just wait .

main thread:

`sample.initalizationSemaphore.acquire();`

in `D3D12Quad::OnInit`:

`sample.initalizationSemaphore.release();`

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/DXResourceBinding.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+#include "pch.h"
+#include "DXResourceBinding.h"
+#include "DXResourceBinding.g.cpp"
+#include "stdafx.h"
+#include "ORTHelpers.h"
+
+using namespace winrt::Microsoft::AI::MachineLearning;
+
+#undef min
```
fdwr, line 12: 🤔 shouldn't be necessary given you have `NOMINMAX` above.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/DXResourceBinding.cpp
```diff
@@ -0,0 +1,59 @@
...
+namespace winrt::WinMLSamplesGalleryNative::implementation
+{
+    // Create ORT Sessions and launch D3D window in a separate thread
+	void DXResourceBinding::LaunchWindow() {
+        // Create ORT Sessions that will be used for preprocessing and classification
+        preprocesingSession = CreateSession(Win32Application::GetAssetPath(L"dx_preprocessor_efficient_net.onnx").c_str());
+        inferenceSession = CreateSession(Win32Application::GetAssetPath(L"efficientnet-lite4-11.onnx").c_str());
+
+        // Spawn the window in a separate thread
+        std::thread d3d_th(Win32Application::Run, &sample, 10);
```
fdwr, line 28: >std::thread [](http://example.com/codeflow?start=8&length=11)

`std::jthread` is generally more robust, because `std::thread` has this braindead default of calling std::terminate upon exiting scope if you don't join or detach (unfortunately it was too late to change, and even the C++ chairman Herb Sutter proposed fixing it)

[https://www.modernescpp.com/index.php/an-improved-thread-with-c-20](https://www.modernescpp.com/index.php/an-improved-thread-with-c-20)

In your case, it doesn't really matter because you're calling `detach` anyway, but for general copy&paste goodness.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/ORTHelpers.cpp
```diff
@@ -0,0 +1,157 @@
...
+    input_node_names.push_back(inputTensorName);
+    std::vector<const char*> output_node_names;
+    output_node_names.push_back(outputTensorName);
+
+    // Evaluate
+    auto output_tensors = session.Run(Ort::RunOptions{ nullptr }, input_node_names.data(), &prev_input, 1, output_node_names.data(), 1);
+    
+    // Get the 1000 EfficientNet classifications as a com_array and return
+    // the results
+    float* floatarr = output_tensors.front().GetTensorMutableData<float>();
```
fdwr, line 110: >floatarr [](http://example.com/codeflow?start=11&length=8)

[nit] With very rare except of really common cases (e.g. ptr), we should spell out whole words.

`floatArray`

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/ORTHelpers.cpp
```diff
@@ -0,0 +1,157 @@
...
+    // Evaluate input (resize from 512 x 512 x 4 to 224 x 224 x 3)
+    Ort::Value outputTensor(nullptr);
+    session.Run(Ort::RunOptions{ nullptr }, input_node_names.data(),
+        &inputTensor, 1, output_node_names.data(), &outputTensor, 1);
+
+    return outputTensor;
+}
+
+// Classify the image using EfficientNet and return the results
+winrt::com_array<float> Eval(Ort::Session& session, const Ort::Value& prev_input) {
```
fdwr, line 96: > { [](http://example.com/codeflow?start=81&length=2)

[nit] inconsistent style vs other functions

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/packages.config
```diff
@@ -1,6 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Microsoft.AI.DirectML" version="1.5.1" targetFramework="native" />
-  <package id="Microsoft.AI.MachineLearning" version="1.9.1" targetFramework="native" />
+  <package id="directxtk12_desktop_2019" version="2022.5.10.1" targetFramework="native" />
+  <package id="Microsoft.AI.DirectML" version="1.8.0" targetFramework="native" />
+  <package id="Microsoft.AI.MachineLearning" version="1.11.0" targetFramework="native" />
+  <package id="Microsoft.ML.OnnxRuntime.DirectML" version="1.11.0" targetFramework="native" />
   <package id="Microsoft.Windows.CppWinRT" version="2.0.210930.14" targetFramework="native" />
```
fdwr, line 7: Consider using the latest:

Microsoft.AI.DirectML.1.9.0
Microsoft.ML.OnnxRuntime.DirectML.1.12.1
...

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
```diff
@@ -0,0 +1,166 @@
...
+}
+
+void Win32Application::CloseWindow()
+{
+    close_window = true;
+}
+
+std::wstring Win32Application::GetAssetPath(LPCWSTR assetName)
+{
+    WCHAR modulePath[512];
```
fdwr, line 151: >512 [](http://example.com/codeflow?start=21&length=3)

Use `MAX_PATH` rather than a hard-coded value.

[https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry)

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
```diff
@@ -0,0 +1,166 @@
...
+    UINT modulePathSize = _countof(modulePath);
+    GetModuleFileName(nullptr, modulePath, modulePathSize);
+
+    const wchar_t* gallery = L"WinMLSamplesGallery";
+    UINT gallerySize = 19; // length of the gallery string
```
fdwr, line 156: >UINT gallerySize = 19 [](http://example.com/codeflow?start=4&length=21)

Consider just using `std::string_view` instead of a hard-coded value (less brittle if you update later). e.g.

```
    constexpr std::wstring_view gallery = L"WinMLSamplesGallery";

    // Erase everything after the first occurence of the gallery string
    std::wstring path(modulePath);
    size_t found = path.find(gallery);
    path.erase(path.begin() + found + gallery.size(), path.end());
```

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+    // Erase everything after the first occurence of the gallery string
+    std::wstring path(modulePath);
+    size_t found = path.find(gallery);
```
fdwr, line 160: >size_t found = path.find(gallery); [](http://example.com/codeflow?start=4&length=34)

If we're certain this can never fail, then an assertion would be smart.

```
size_t offset = path.find(gallery);
assert(offset != std::string::npos);
```

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
```diff
@@ -0,0 +1,166 @@
...
+    path.erase(path.begin() + found + gallerySize, path.end());
+
+    std::wstring native = L"\\WinMLSamplesGalleryNative\\";
```
fdwr, line 163: >std::wstring native = L"\\WinMLSamplesGalleryNative\\"; [](http://example.com/codeflow?start=4&length=55)

std::wstring_view works better here - no need to assign this as a temporary string allocation since it never gets modified.

```
std::wstring_view galleryPathNative = L"\\WinMLSamplesGalleryNative\\";
```

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// Main message handler for the sample.
+LRESULT CALLBACK Win32Application::WindowProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
+{
+    D3D12Quad* pSample = reinterpret_cast<D3D12Quad*>(GetWindowLongPtr(hWnd, GWLP_USERDATA));
+
+    switch (message)
+    {
+    case WM_CREATE:
```
fdwr, line 116: >WM_CREATE [](http://example.com/codeflow?start=9&length=9)

WM_NCCREATE is safer here for setting the class window pointer because it's the first message you receive, but there can be other messages before WM_CREATE. (In this case, it's actually okay because we don't have a bunch of other messages to handle, but better safe than sorry and better to set right example for others)

```
    case WM_NCCREATE:
        {
            // Save the D3D12Quad* passed in to CreateWindow.
            LPCREATESTRUCT pCreateStruct = reinterpret_cast<LPCREATESTRUCT>(lParam);
            SetWindowLongPtr(hWnd, GWLP_USERDATA, reinterpret_cast<LONG_PTR>(pCreateStruct->lpCreateParams));
            return DefWindowProc(hWnd, message, wParam, lParam);
        }
```

Note DefWindowProc should be called for WM_NCCREATE.

[https://devblogs.microsoft.com/oldnewthing/20191014-00/?p=10299](https://devblogs.microsoft.com/oldnewthing/20191014-00/?p=10299)

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/DXResourceBinding.cpp
nums11: Do you mean to separate the comment and the line with the function definition with a new line?

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGallery/Pages/SampleBasePage.xaml.cs
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGallery/Samples/DXResourceBindingORT/DXResourceBindingORT.xaml.cs
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGallery/Samples/DXResourceBindingORT/DXResourceBindingORT.xaml.cs
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.h
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.h
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
nums11: Nope. Removing comment.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/DXResourceBinding.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/DXResourceBinding.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/ORTHelpers.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/ORTHelpers.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/DXResourceBinding.cpp
```diff
@@ -0,0 +1,59 @@
...
+static std::optional<Ort::Session> preprocesingSession;
+static std::optional<Ort::Session> inferenceSession;
+D3D12Quad sample(800, 600, L"D3D12 Quad");
+
+namespace winrt::WinMLSamplesGalleryNative::implementation
+{
+    // Create ORT Sessions and launch D3D window in a separate thread
+	void DXResourceBinding::LaunchWindow() {
+        // Create ORT Sessions that will be used for preprocessing and classification
+        preprocesingSession = CreateSession(Win32Application::GetAssetPath(L"dx_preprocessor_efficient_net.onnx").c_str());
```
linnealovespie, line 22: >dx_preprocessor_efficient_net [](http://example.com/codeflow?start=77&length=29)

Did you keep the code for LearningModelBuilder anywhere, or is it just the onnx file?

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/DXResourceBinding.cpp
```diff
@@ -0,0 +1,59 @@
...
+        inferenceSession = CreateSession(Win32Application::GetAssetPath(L"efficientnet-lite4-11.onnx").c_str());
```
linnealovespie, line 23: I can't find a good way to comment on the file itself, but it looks like there's a version of this model in WinMLSamplesGalleryNative as well as in WinMLSamplesGallery/Models. Are you using both or is one redundant?

Samples/WinMLSamplesGallery/WinMLSamplesGallery/SampleMetadata/SampleMetadata.json
```diff
@@ -82,6 +82,17 @@
...
       "Docs": [],
       "IsRecentlyAdded": true
+    },
+    {
+      "Title": "DX Resource Binding in ORT",
+      "DescriptionShort": "Learn how to bind and evaluate DirectX Resources using ONNX Runtime.",
+      "Description": "The sample showcases how to create and bind OrtValue objects from ID3D12 Resources using the ONNX Runtime C-API.",
+      "Icon": "\uE155",
+      "Tag": "DXResourceBindingORT",
+      "XAMLGithubLink": "https://github.com/microsoft/Windows-Machine-Learning/blob/master/Samples/WinMLSamplesGallery/WinMLSamplesGallery/Samples/AdapterSelection/AdapterSelection.xaml",
```
linnealovespie, line 92: >AdapterSelection [](http://example.com/codeflow?start=164&length=16)

Note: I've got to remember to do it as well, but update this to be DX resource xaml page instead.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/WinMLSamplesGalleryNative.vcxproj
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/WinMLSamplesGalleryNative.vcxproj
nums11: Yes, removed.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/packages.config
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/DXResourceBinding.cpp
nums11: Just the onnx file

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/D3D12Quad.h
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/DXResourceBinding.cpp
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/DXResourceBinding.cpp
nums11: Both are in use. The reason I copied it into the native project as well is because the GetAssetPath function starts searching for assets in the native project. I could use the version of the model in use in the front end, but I'm not sure if it's good practice to reach across projects.

Samples/WinMLSamplesGallery/WinMLSamplesGallery/SampleMetadata/SampleMetadata.json
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/Win32Application.cpp
nums11: Done.

linnealovespie (approved):

PR #491, opened 2022-07-13
Set Model Names in Samples for Telemetry

nums11:
This PR updates the Samples Gallery to use the SetName experimental API to set model names inside samples. The updated model name contains the name of the sample in which the model was used (e.g. "SqueezeNet_ImageClassifier") so that when telemetry is sent during session creation, the usage of specific samples will be tracked.

- Samples Gallery is updated to use Microsoft.AI.MachineLearrning 1.11
- Uses SetName Experimental API: https://github.com/microsoft/onnxruntime/pull/10518

### DIFF
--- a/Samples/WinMLSamplesGallery/WinMLSamplesGallery/Pages/SampleBasePage.xaml.cs
+++ b/Samples/WinMLSamplesGallery/WinMLSamplesGallery/Pages/SampleBasePage.xaml.cs
@@ -1,6 +1,8 @@
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Navigation;
+using Microsoft.AI.MachineLearning;
+using Microsoft.AI.MachineLearning.Experimental;
 using System;
 
 namespace WinMLSamplesGallery
@@ -55,6 +57,13 @@ namespace WinMLSamplesGallery
                 var page = (Samples.Batching)SampleFrame.Content;
                 page.StopAllEvents();
             }
+        }
+
+        public static void SetModelNameForTelemetry(String modelName, String sampleName, LearningModel model)
+        {
+            var telemetryModelName = modelName + "_" + sampleName;
+            var experimentalModel = new LearningModelExperimental(model);
+            experimentalModel.SetName(telemetryModelName);
         }
     }
 }
--- a/Samples/WinMLSamplesGallery/WinMLSamplesGallery/Samples/AdapterSelection/AdapterSelection.xaml.cs
+++ b/Samples/WinMLSamplesGallery/WinMLSamplesGallery/Samples/AdapterSelection/AdapterSelection.xaml.cs
@@ -34,6 +34,7 @@ namespace WinMLSamplesGallery.Samples
             adapter_options.AddRange(adapters);
             AdapterListView.ItemsSource = adapter_options;
 
+            // Run in background thread to avoid blocking UI on sample launch
             Task.Run(() => SetModelNameForTelemetry(device));
         }
 
@@ -103,6 +104,7 @@ namespace WinMLSamplesGallery.Samples
             return adapters;
         }
 
+        // Session must be created for telemety to be sent
         private void SetModelNameForTelemetry(LearningModelDevice device)
         {
             var modelName = "squeezenet1.1-7.onnx";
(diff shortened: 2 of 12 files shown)


Samples/WinMLSamplesGallery/WinMLSamplesGallery/Samples/AdapterSelection/AdapterSelection.xaml.cs
```diff
@@ -33,6 +33,8 @@ public AdapterSelection()
 
             adapter_options.AddRange(adapters);
             AdapterListView.ItemsSource = adapter_options;
+
+            Task.Run(() => SetModelNameForTelemetry(device));
```
martinb35, line 37: why can't you just call setmodelnamefortelemetry directly?

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/WinMLSamplesGalleryNative.vcxproj
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\packages\Microsoft.AI.MachineLearning.1.9.1\build\native\Microsoft.AI.MachineLearning.props" Condition="Exists('..\packages\Microsoft.AI.MachineLearning.1.9.1\build\native\Microsoft.AI.MachineLearning.props')" />
-  <Import Project="..\packages\Microsoft.AI.DirectML.1.5.1\build\Microsoft.AI.DirectML.props" Condition="Exists('..\packages\Microsoft.AI.DirectML.1.5.1\build\Microsoft.AI.DirectML.props')" />
+  <Import Project="..\packages\Microsoft.AI.MachineLearning.1.11.0\build\native\Microsoft.AI.MachineLearning.props" Condition="Exists('..\packages\Microsoft.AI.MachineLearning.1.11.0\build\native\Microsoft.AI.MachineLearning.props')" />
+  <Import Project="..\packages\Microsoft.AI.DirectML.1.8.0\build\Microsoft.AI.DirectML.props" Condition="Exists('..\packages\Microsoft.AI.DirectML.1.8.0\build\Microsoft.AI.DirectML.props')" />
```
martinb35, line 4: I believe ORT 1.11 went out with dml 1.8.2 - can you use that instead?

Samples/WinMLSamplesGallery/WinMLSamplesGallery/Samples/AdapterSelection/AdapterSelection.xaml.cs
nums11: It blocks the UI thread and causes the sample to take a while to load since this is called in the constructor. Just added a comment to address this.

Samples/WinMLSamplesGallery/WinMLSamplesGalleryNative/WinMLSamplesGalleryNative.vcxproj
nums11: Done.

Samples/WinMLSamplesGallery/WinMLSamplesGallery/Samples/AdapterSelection/AdapterSelection.xaml.cs
linnealovespie: why is it different only for AdapterSelection page? looks like the rest of them only call SampleBasePage.SetModelNameForTelemetry

Samples/WinMLSamplesGallery/WinMLSamplesGallery/Pages/SampleBasePage.xaml.cs
```diff
@@ -56,5 +58,12 @@ protected override async void OnNavigatingFrom(NavigatingCancelEventArgs e)
                 page.StopAllEvents();
             }
         }
+
+        public static void SetModelNameForTelemetry(String modelName, String sampleName, LearningModel model)
```
linnealovespie, line 62: >LearningModel model [](http://example.com/codeflow?start=89&length=19)

looks like you need access to the model from c# land. the streaming demo manages the model from the native side, is there a way to get the same telemetry without the specific model instance?

martinb35 (approved):

PR #508, opened 2022-12-09
User/brianma/stringbinding

martinb35:
we had a request for string input support, so this change adds basic functionality. Also, updated the NuGet packages to the latest versions.

### DIFF
--- a/Tools/WinMLRunner/src/BindingUtilities.cpp
+++ b/Tools/WinMLRunner/src/BindingUtilities.cpp
@@ -646,6 +646,28 @@ namespace BindingUtilities
         }
     }
 
+    TensorString CreateTensorString(const std::vector<int64_t>& tensorShape)
+    {
+        auto length = 1;
+        for (auto&& dim : tensorShape)
+        {
+            if (dim > 0)
+            {
+                length *= dim;
+            }
+        }
+
+        std::vector<winrt::hstring> strings;
+
+        for (auto i = 0; i < length; ++i)
+        {
+            auto str = to_hstring(i);
+            strings.push_back(str);
+        }
+
+        return TensorString::CreateFromArray(tensorShape, strings);
+    }
+
     // Process the descriptor to gather and normalize the shape
     void ProcessDescriptor(const ILearningModelFeatureDescriptor& description, std::vector<int64_t>& shape,
                            TensorKind& tensorKind, InputBufferDesc& inputBufferDesc)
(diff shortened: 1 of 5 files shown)


nums11 (approved):

Tools/WinMLRunner/src/BindingUtilities.cpp
```diff
@@ -646,6 +646,28 @@ namespace BindingUtilities
...
+            {
+                length *= dim;
+            }
+        }
+
+        std::vector<winrt::hstring> strings;
+
+        for (auto i = 0; i < length; ++i)
+        {
+            auto str = to_hstring(i);
```
PatriceVignola, line 664: I'm not sure I understand what this logic is supposed to do. Why do we want to create a list of strings that looks like `{"0", "1", "2", ..., "100", "101", "102"}`?

Tools/WinMLRunner/src/BindingUtilities.cpp
```diff
@@ -646,6 +646,28 @@ namespace BindingUtilities
         }
     }
 
+    TensorString CreateTensorString(const std::vector<int64_t>& tensorShape)
+    {
+        auto length = 1;
+        for (auto&& dim : tensorShape)
+        {
+            if (dim > 0)
```
PatriceVignola, line 654: Is this the same behavior as the other types of tensors? It feels like we should either fail early or keep the 0's in the multiplications (empty tensors are valid in ORT, although it wouldn't be very useful for an input). Ignoring the 0 would give us a "valid" shape, but probably not what the caller intended.

Tools/WinMLRunner/src/BindingUtilities.cpp
martinb35: it's just a way to create valid strings. They don't have any meaning, but they are unique, valid strings to pass to the model. Basically garbage data like the other garbage data generating methods.

Tools/WinMLRunner/src/BindingUtilities.cpp
martinb35: Are there use cases for 0 size tensors? I don't think I've seen those. This is mostly to catch when a model has "-1" to symbolize a free dimension, but multiplying by -1 would obviously cause problems in the size calculation. In those cases, we just pick "1" as the free dimension since it's impossible to know what the actual use case is in a generic test tool.

PatriceVignola (approved):

Tools/WinMLRunner/src/BindingUtilities.cpp
fdwr: Already submitted, but I want to dispel the myth that these two things are invalid:

- 0 size tensors are valid - they're simply empty tensors. In most cases, operator execution on them is a nop (just like `memset(p, 0, 0)` being a nop), but they can be the output of a slice that sliced away everything, and they can be the input to a padding operation (which then yields a non-empty tensor).
- 0D tensors are valid - they're just scalar values.

PR #511, opened 2023-01-23
Update Yolov4 sample to use the NuGet

nums11:
- This PR updates the Yolov4 sample to use NuGet WinML instead of inbox, which fixes the issue for the sample not working on Windows 10.

### DIFF
--- a/Samples/Tutorial Samples/YOLOv4ObjectDetection/YOLOv4ObjectDetection/Model.cs
+++ b/Samples/Tutorial Samples/YOLOv4ObjectDetection/YOLOv4ObjectDetection/Model.cs
@@ -4,10 +4,10 @@ using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
-using Windows.AI.MachineLearning;
 using Windows.Media;
 using Windows.Storage;
 using Windows.UI.Xaml;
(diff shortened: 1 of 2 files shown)


Samples/Tutorial Samples/YOLOv4ObjectDetection/YOLOv4ObjectDetection/Model.cs
```diff
@@ -4,10 +4,10 @@
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
-using Windows.AI.MachineLearning;
```
martinb35, line 7: can you leave this in but commented out with a comment that says, "'Windows.AI.MachineLearning' refers to the libraries included in Windows, while 'Microsoft.AI.MachineLearning' refers to the externally available NuGet package." or something like that. what do you think? It would be good for people to know how to use either.

martinb35 (approved):

Samples/Tutorial Samples/YOLOv4ObjectDetection/YOLOv4ObjectDetection/Model.cs
nums11: Yea that makes sense.

